PR #34, opened 2025-09-04
[DPE-7322] Support predefined roles

sinclert-canonical:
This PR supports the set of predefined roles proposed in [this specification](https://docs.google.com/document/d/1U1tAWx9a7eb0HvdiS7lnSJS8oHqcl6F9PnAoRdcMjmc/edit?tab=t.0).

Contents have been ported from the MySQL Router VM and K8s approved PRs. However, clauses to create a DBA role for each created databases have been added (see [code](https://github.com/canonical/mysql-router-operators/blob/c0ce9031d886f70894cbc168efc1082d6e7e95b5/common/common/mysql_shell/__init__.py#L159-L165)), which is something that got added late into the spec review process.

### DIFF
--- a/common/common/mysql_shell/__init__.py
+++ b/common/common/mysql_shell/__init__.py
@@ -13,7 +13,6 @@ import pathlib
 import typing
 
 import jinja2
-import tenacity
 
 from .. import container, server_exceptions, utils
 
@@ -181,13 +180,12 @@ class Shell:
         if database in self._get_mysql_databases():
             return
 
-        rolename = self._build_application_database_dba_role(database)
-
+        role_name = self._build_application_database_dba_role(database)
         statements = [
-            f"CREATE ROLE `{rolename}`",
+            f"CREATE ROLE `{role_name}`",
             f"CREATE DATABASE `{database}`",
-            f"GRANT SELECT, INSERT, DELETE, UPDATE, EXECUTE ON `{database}`.* TO {rolename}",
-            f"GRANT ALTER, ALTER ROUTINE, CREATE, CREATE ROUTINE, CREATE VIEW, DROP, INDEX, LOCK TABLES, REFERENCES, TRIGGER ON `{database}`.* TO {rolename}",
+            f"GRANT SELECT, INSERT, DELETE, UPDATE, EXECUTE ON `{database}`.* TO {role_name}",
+            f"GRANT ALTER, ALTER ROUTINE, CREATE, CREATE ROUTINE, CREATE VIEW, DROP, INDEX, LOCK TABLES, REFERENCES, TRIGGER ON `{database}`.* TO {role_name}",
         ]
 
         mysql_roles = self._get_mysql_roles("charmed_%")
@@ -218,15 +216,7 @@ class Shell:
 
     def create_application_database(self, *, database: str, username: str) -> str:
         """Create both the database and the relation user, returning its password."""
-        for attempt in tenacity.Retrying(
-            retry=tenacity.retry_if_exception_type(ShellDBError),
-            reraise=True,
-            stop=tenacity.stop_after_delay(30),
-            wait=tenacity.wait_fixed(5),
-        ):
-            with attempt:
-                self._create_application_database(database=database)
-
+        self._create_application_database(database=database)
         return self._create_application_user(database=database, username=username)
 
     def add_attributes_to_mysql_router_user(
--- a/common/common/mysql_shell/__init__.py
+++ b/common/common/mysql_shell/__init__.py
@@ -19,6 +19,12 @@ from .. import container, server_exceptions, utils
 if typing.TYPE_CHECKING:
     from ..relations import database_requires
 
+_ROLE_DML = "charmed_dml"
+_ROLE_READ = "charmed_read"
+
+ROLE_DBA_PREFIX = "charmed_dba"
+ROLE_MAX_LENGTH = 32
+
 logger = logging.getLogger(__name__)
 
 
@@ -123,17 +129,67 @@ class Shell:
             attributes.update(additional_attributes)
         return json.dumps(attributes)
 
-    def create_application_database_and_user(self, *, username: str, database: str) -> str:
-        """Create database and user for related database_provides application."""
-        attributes = self._get_attributes()
-        logger.debug(f"Creating {database=} and {username=} with {attributes=}")
-        password = utils.generate_password()
-        self._run_sql([
+    # TODO python3.10 min version: Use `set` instead of `typing.Set`
+    def _get_mysql_roles(self, name_pattern: str) -> typing.Set[str]:
+        """Returns a set with the MySQL roles."""
+        logger.debug(f"Getting MySQL roles with {name_pattern=}")
+        output_file = self._container.path("/tmp/mysqlsh_output.json")
+        self._run_code(
+            _jinja_env.get_template("get_mysql_roles_with_pattern.py.jinja").render(
+                name_pattern=name_pattern,
+                output_filepath=output_file.relative_to_container,
+            )
+        )
+        with output_file.open("r") as file:
+            rows = json.load(file)
+        output_file.unlink()
+        logger.debug(f"MySQL roles found for {name_pattern=}: {len(rows)}")
+        return {row[0] for row in rows}
+
+    def _create_application_database(self, *, database: str, rolename: str) -> str:
+        """Create database for related database_provides application."""
+        statements = [
             f"CREATE DATABASE IF NOT EXISTS `{database}`",
+            f"CREATE ROLE IF NOT EXISTS `{rolename}`",
+            f"GRANT SELECT, INSERT, DELETE, UPDATE, EXECUTE ON `{database}`.* TO {rolename}",
+            f"GRANT ALTER, ALTER ROUTINE, CREATE, CREATE ROUTINE, CREATE VIEW, DROP, INDEX, LOCK TABLES, REFERENCES, TRIGGER ON `{database}`.* TO {rolename}",
+        ]
+
+        mysql_roles = self._get_mysql_roles("charmed_%")
+        if _ROLE_READ in mysql_roles:
+            statements.append(
+                f"GRANT SELECT ON `{database}`.* TO {_ROLE_READ}",
+            )
+        if _ROLE_DML in mysql_roles:
+            statements.append(
+                f"GRANT SELECT, INSERT, DELETE, UPDATE ON `{database}`.* TO {_ROLE_DML}",
+            )
+
+        logger.debug(f"Creating {database=}")
+        self._run_sql(statements)
+        logger.debug(f"Created {database=}")
+        return database
+
+    def _create_application_user(self, *, database: str, username: str) -> str:
+        """Create database user for related database_provides application."""
+        attributes = self._get_attributes()
+        password = utils.generate_password()
+        logger.debug(f"Creating {username=} with {attributes=}")
+        self._run_sql([
             f"CREATE USER `{username}` IDENTIFIED BY '{password}' ATTRIBUTE '{attributes}'",
             f"GRANT ALL PRIVILEGES ON `{database}`.* TO `{username}`",
         ])
-        logger.debug(f"Created {database=} and {username=} with {attributes=}")
+        logger.debug(f"Created {username=} with {attributes=}")
+        return password
+
+    def create_application_database(self, *, database: str, username: str) -> str:
+        """Create both the database and the relation user, returning its password."""
+        rolename = f"{ROLE_DBA_PREFIX}_{database}"
+        if len(rolename) >= ROLE_MAX_LENGTH:
+            raise ValueError("Database DBA role longer than 32 characters")
+
+        ________ = self._create_application_database(database=database, rolename=rolename)
+        password = self._create_application_user(database=database, username=username)
         return password
 
     def add_attributes_to_mysql_router_user(
--- a/common/common/mysql_shell/__init__.py
+++ b/common/common/mysql_shell/__init__.py
@@ -13,6 +13,7 @@ import pathlib
 import typing
 
 import jinja2
+import tenacity
 
 from .. import container, server_exceptions, utils
 
@@ -21,9 +22,7 @@ if typing.TYPE_CHECKING:
 
 _ROLE_DML = "charmed_dml"
 _ROLE_READ = "charmed_read"
-
-ROLE_DBA_PREFIX = "charmed_dba"
-ROLE_MAX_LENGTH = 32
+_ROLE_MAX_LENGTH = 32
 
 logger = logging.getLogger(__name__)
 
@@ -130,6 +129,22 @@ class Shell:
         return json.dumps(attributes)
 
     # TODO python3.10 min version: Use `set` instead of `typing.Set`
+    def _get_mysql_databases(self) -> typing.Set[str]:
+        """Returns a set with the MySQL databases."""
+        logger.debug(f"Getting MySQL databases")
+        output_file = self._container.path("/tmp/mysqlsh_output.json")
+        self._run_code(
+            _jinja_env.get_template("get_mysql_databases.py.jinja").render(
+                output_filepath=output_file.relative_to_container,
+            )
+        )
+        with output_file.open("r") as file:
+            rows = json.load(file)
+        output_file.unlink()
+        logger.debug(f"MySQL databases found: {len(rows)}")
+        return {row[0] for row in rows}
+
+    # TODO python3.10 min version: Use `set` instead of `typing.Set`
     def _get_mysql_roles(self, name_pattern: str) -> typing.Set[str]:
         """Returns a set with the MySQL roles."""
         logger.debug(f"Getting MySQL roles with {name_pattern=}")
@@ -146,11 +161,31 @@ class Shell:
         logger.debug(f"MySQL roles found for {name_pattern=}: {len(rows)}")
         return {row[0] for row in rows}
 
-    def _create_application_database(self, *, database: str, rolename: str) -> str:
+    def _build_application_database_dba_role(self, database: str) -> str:
+        """Builds the database-level DBA role, given length constraints."""
+        role_prefix = "charmed_dba"
+        role_suffix = "XX"
+
+        role_name_available = _ROLE_MAX_LENGTH - len(role_prefix) - len(role_suffix) - 2
+        role_name_description = database[:role_name_available]
+        role_name_collisions = self._get_mysql_roles(f"{role_prefix}_{role_name_description}_%")
+
+        return "_".join((
+            role_prefix,
+            role_name_description,
+            str(len(role_name_collisions)).zfill(len(role_suffix)),
+        ))
+
+    def _create_application_database(self, *, database: str) -> None:
         """Create database for related database_provides application."""
+        if database in self._get_mysql_databases():
+            return
+
+        rolename = self._build_application_database_dba_role(database)
+
         statements = [
-            f"CREATE DATABASE IF NOT EXISTS `{database}`",
-            f"CREATE ROLE IF NOT EXISTS `{rolename}`",
+            f"CREATE ROLE `{rolename}`",
+            f"CREATE DATABASE `{database}`",
             f"GRANT SELECT, INSERT, DELETE, UPDATE, EXECUTE ON `{database}`.* TO {rolename}",
             f"GRANT ALTER, ALTER ROUTINE, CREATE, CREATE ROUTINE, CREATE VIEW, DROP, INDEX, LOCK TABLES, REFERENCES, TRIGGER ON `{database}`.* TO {rolename}",
         ]
@@ -168,7 +203,6 @@ class Shell:
         logger.debug(f"Creating {database=}")
         self._run_sql(statements)
         logger.debug(f"Created {database=}")
-        return database
 
     def _create_application_user(self, *, database: str, username: str) -> str:
         """Create database user for related database_provides application."""
@@ -184,13 +218,16 @@ class Shell:
 
     def create_application_database(self, *, database: str, username: str) -> str:
         """Create both the database and the relation user, returning its password."""
-        rolename = f"{ROLE_DBA_PREFIX}_{database}"
-        if len(rolename) >= ROLE_MAX_LENGTH:
-            raise ValueError("Database DBA role longer than 32 characters")
+        for attempt in tenacity.Retrying(
+            retry=tenacity.retry_if_exception_type(ShellDBError),
+            reraise=True,
+            stop=tenacity.stop_after_delay(30),
+            wait=tenacity.wait_fixed(5),
+        ):
+            with attempt:
+                self._create_application_database(database=database)
 
-        ________ = self._create_application_database(database=database, rolename=rolename)
-        password = self._create_application_user(database=database, username=username)
-        return password
+        return self._create_application_user(database=database, username=username)
 
     def add_attributes_to_mysql_router_user(
         self, *, username: str, router_id: str, unit_name: str
--- a/common/common/mysql_shell/templates/get_mysql_databases.py.jinja
+++ b/common/common/mysql_shell/templates/get_mysql_databases.py.jinja
@@ -1,0 +1,10 @@
+import json
+
+result = session.run_sql("SHOW DATABASES")
+rows = result.fetch_all()
+# mysqlsh objects are weird—they quack (i.e. duck typing) like standard Python objects (e.g. list,
+# dict), but do not serialize to JSON correctly.
+# Cast to str & load from JSON str before serializing
+rows = json.loads(str(rows))
+with open("{{ output_filepath }}", "w") as file:
+    json.dump(rows, file)
--- a/common/common/relations/database_provides.py
+++ b/common/common/relations/database_provides.py
@@ -30,23 +30,10 @@ class _UnsupportedExtraUserRole(status_exception.StatusException):
 
     def __init__(self, *, app_name: str, endpoint_name: str) -> None:
         message = (
-            f"{app_name} app requested unsupported extra user role on "
-            f"{endpoint_name} endpoint"
+            f"{app_name} app requested unsupported extra user role on {endpoint_name} endpoint"
         )
         logger.warning(message)
         super().__init__(ops.BlockedStatus(message))
-
-
-class _InvalidDatabaseName(status_exception.StatusException):
-    """Application charm requested an invalid database name"""
-
-    def __init__(self, *, app_name: str, endpoint_name: str, exception_msg: str) -> None:
-        message = (
-            f"{app_name} app requested an invalid database name on "
-            f"{endpoint_name} endpoint: {exception_msg}"
-        )
-        logger.warning(message)
-        super().__init__(ops.BlockedStatus(exception_msg))
 
 
 class _Relation:
@@ -92,19 +79,9 @@ class _RelationThatRequestedUser(_Relation):
         if isinstance(event, ops.RelationBrokenEvent) and event.relation.id == self._id:
             raise _RelationBreaking
         self._database: str = self._databag["database"]
-        self._app_name: str = relation.app.name
-        self._endpoint_name: str = relation.name
-        database_dba_role = f"{mysql_shell.ROLE_DBA_PREFIX}_{self._database}"
-        if len(database_dba_role) >= mysql_shell.ROLE_MAX_LENGTH:
-            raise _InvalidDatabaseName(
-                app_name=self._app_name,
-                endpoint_name=self._endpoint_name,
-                exception_msg="Database DBA role longer than 32 characters",
-            )
         if self._databag.get("extra-user-roles"):
             raise _UnsupportedExtraUserRole(
-                app_name=self._app_name,
-                endpoint_name=self._endpoint_name,
+                app_name=relation.app.name, endpoint_name=relation.name
             )
 
     def _set_databag(
--- a/common/common/relations/database_provides.py
+++ b/common/common/relations/database_provides.py
@@ -30,10 +30,23 @@ class _UnsupportedExtraUserRole(status_exception.StatusException):
 
     def __init__(self, *, app_name: str, endpoint_name: str) -> None:
         message = (
-            f"{app_name} app requested unsupported extra user role on {endpoint_name} endpoint"
+            f"{app_name} app requested unsupported extra user role on "
+            f"{endpoint_name} endpoint"
         )
         logger.warning(message)
         super().__init__(ops.BlockedStatus(message))
+
+
+class _InvalidDatabaseName(status_exception.StatusException):
+    """Application charm requested an invalid database name"""
+
+    def __init__(self, *, app_name: str, endpoint_name: str, exception_msg: str) -> None:
+        message = (
+            f"{app_name} app requested an invalid database name on "
+            f"{endpoint_name} endpoint: {exception_msg}"
+        )
+        logger.warning(message)
+        super().__init__(ops.BlockedStatus(exception_msg))
 
 
 class _Relation:
@@ -79,9 +92,19 @@ class _RelationThatRequestedUser(_Relation):
         if isinstance(event, ops.RelationBrokenEvent) and event.relation.id == self._id:
             raise _RelationBreaking
         self._database: str = self._databag["database"]
+        self._app_name: str = relation.app.name
+        self._endpoint_name: str = relation.name
+        database_dba_role = f"{mysql_shell.ROLE_DBA_PREFIX}_{self._database}"
+        if len(database_dba_role) >= mysql_shell.ROLE_MAX_LENGTH:
+            raise _InvalidDatabaseName(
+                app_name=self._app_name,
+                endpoint_name=self._endpoint_name,
+                exception_msg="Database DBA role longer than 32 characters",
+            )
         if self._databag.get("extra-user-roles"):
             raise _UnsupportedExtraUserRole(
-                app_name=relation.app.name, endpoint_name=relation.name
+                app_name=self._app_name,
+                endpoint_name=self._endpoint_name,
             )
 
     def _set_databag(
@@ -121,10 +144,10 @@ class _RelationThatRequestedUser(_Relation):
         shell.delete_user(username, must_exist=False)
         logger.debug("Deleted user if exists before creating user")
 
-        password = shell.create_application_database_and_user(
-            username=username, database=self._database
+        password = shell.create_application_database(
+            database=self._database,
+            username=username,
         )
-
         self._set_databag(
             username=username,
             password=password,
--- a/common/common/mysql_shell/templates/get_mysql_roles_with_pattern.py.jinja
+++ b/common/common/mysql_shell/templates/get_mysql_roles_with_pattern.py.jinja
@@ -1,0 +1,12 @@
+import json
+
+result = session.run_sql(
+    "SELECT user FROM mysql.user WHERE user LIKE '{{ name_pattern }}'"
+)
+rows = result.fetch_all()
+# mysqlsh objects are weird—they quack (i.e. duck typing) like standard Python objects (e.g. list,
+# dict), but do not serialize to JSON correctly.
+# Cast to str & load from JSON str before serializing
+rows = json.loads(str(rows))
+with open("{{ output_filepath }}", "w") as file:
+    json.dump(rows, file)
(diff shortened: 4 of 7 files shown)


common/common/mysql_shell/__init__.py
```diff
@@ -123,17 +127,62 @@ def _get_attributes(self, additional_attributes: dict = None) -> str:
             attributes.update(additional_attributes)
         return json.dumps(attributes)
 
-    def create_application_database_and_user(self, *, username: str, database: str) -> str:
-        """Create database and user for related database_provides application."""
```
carlcsaposs-canonical, line 127: same as https://github.com/canonical/mysql-router-operator/pull/269#discussion_r2269282423

common/common/mysql_shell/__init__.py
```diff
@@ -123,17 +127,62 @@ def _get_attributes(self, additional_attributes: dict = None) -> str:
...
+        output_file.unlink()
+        logger.debug(f"MySQL roles found for {name_pattern=}: {len(rows)}")
+        return {row[0] for row in rows}
+
+    def create_application_database(self, *, database: str) -> str:
+        """Create database for related database_provides application."""
+        role_name = f"charmed_dba_{database}"
+        if len(role_name) >= _ROLE_MAX_LENGTH:
+            logger.exception("Failed to create application database")
+            raise ValueError("Role name longer than 32 characters")
```
carlcsaposs-canonical: - did we update charm-relation-interfaces to include a length limit on the database name?
- also might be worth updating message to say database name is too long (to be more relevant to the input that's controlled by the app requesting the database & potentially the user)
- also we should use blocked status instead of raising an exception since charm should recover if relation changes to a valid database name—if we raise, we can never recover since we won't get updates to relation databag. See invalid extra-user-roles handling for an example

common/common/mysql_shell/__init__.py
```diff
@@ -123,17 +127,62 @@ def _get_attributes(self, additional_attributes: dict = None) -> str:
...
+
+        statements = [
+            f"CREATE DATABASE IF NOT EXISTS `{database}`",
+            f"CREATE ROLE IF NOT EXISTS `{role_name}`",
+            f"GRANT SELECT, INSERT, DELETE, UPDATE, EXECUTE ON `{database}`.* TO {role_name}",
+            f"GRANT ALTER, ALTER ROUTINE, CREATE, CREATE ROUTINE, CREATE VIEW, DROP, INDEX, LOCK TABLES, REFERENCES, TRIGGER ON `{database}`.* TO {role_name}",
```
carlcsaposs-canonical: can these be combined into one grant?

carlcsaposs-canonical (changes requested):

common/common/mysql_shell/__init__.py
sinclert-canonical: Same as https://github.com/canonical/mysql-router-operator/pull/269#discussion_r2313411480.

I am not going to propose a 2-line function for the sake of if. If you prefer to have a single exposed function to create new database resources / roles, I will move role-name length validation there too.

common/common/mysql_shell/__init__.py
sinclert-canonical: > Did we update charm-relation-interfaces to include a length limit on the database name?

No. The 32 character limit is highly coupled with the MySQL pre-defined roles spec, and something that cannot be generalized for all consumers of the database interface (MySQL, PostgreSQL, MongoDB...). That validation needs to happen on MySQL Server / MySQL Router side.

> Might be worth updating message to say database name is too long

Will do.

>  We should use blocked status instead of raising an exception

Will do.

common/common/mysql_shell/__init__.py
sinclert-canonical: They could, but I would rather not to.

The 1st set of permissions target _data-related_ operations, while the 2nd set target _schema-related_ operations (table, view, index, foreign-keys...). I already left them separated in MySQL operators (see [VM](https://github.com/canonical/mysql-operator/blob/5129fa8d1c358ce7f9b158890d9bc4c36a114a30/lib/charms/mysql/v0/mysql.py#L1523-L1527) and [K8s](https://github.com/canonical/mysql-k8s-operator/blob/26335553a57e9f95fbbab99c0a24d53a224eb649/lib/charms/mysql/v0/mysql.py#L1523-L1527)), and I think it makes it easier to scan privileges this way.

common/common/mysql_shell/__init__.py
carlcsaposs-canonical: yes, I would prefer one method (https://github.com/canonical/mysql-router-operator/pull/269#discussion_r2313653383) to keep the mysql_shell interface the same

common/common/mysql_shell/__init__.py
carlcsaposs-canonical: > for all consumers of the database interface

I think it would apply to all users of the mysql_client interface

https://github.com/canonical/charm-relation-interfaces/tree/main/interfaces/mysql_client/v0

common/common/mysql_shell/__init__.py
carlcsaposs-canonical: would there be any issues if one of the grants succeeds and the other fails?
was thinking it could be beneficial (e.g. for debugging) to make the operation more atomic

common/common/mysql_shell/__init__.py
sinclert-canonical: Done in commit https://github.com/canonical/mysql-router-operators/pull/34/commits/7e4758bf34a3d954fd743115fd25969de3de9acb.

common/common/mysql_shell/__init__.py
sinclert-canonical: I thought you were referring to **all** Data Platform _database_ type of interfaces. MySQL could be updated, yes.

Code suggestions applied in commit https://github.com/canonical/mysql-router-operators/pull/34/commits/7e4758bf34a3d954fd743115fd25969de3de9acb.

common/common/mysql_shell/__init__.py
sinclert-canonical: > Would there be any issues if one of the grants succeeds and the other fails?

Sure, but that highly depends on the SQL execution strategy each repo follows.

In the specific case of MySQL Router, we are not wrapping every SQL execution into a transaction [here](https://github.com/canonical/mysql-router-operators/blob/5da2b8bb1d4752430fbd0641a2f5941f1ed5f21d/common/common/mysql_shell/templates/run_sql.py.jinja), so there are no guarantees that all statements will be executed atomically. Wrapping everything in a transaction would slightly change behavior across every run SQL set of statements, something worth exploring in a separate effort.

sinclert-canonical: > Would there be any issues if one of the grants succeeds and the other fails?

Sure, but that highly depends on the SQL execution strategy each repo follows.

In the specific case of MySQL Router, we are not wrapping every SQL execution into a transaction [here](https://github.com/canonical/mysql-router-operators/blob/5da2b8bb1d4752430fbd0641a2f5941f1ed5f21d/common/common/mysql_shell/templates/run_sql.py.jinja), so there are no guarantees that all statements will be executed atomically. Wrapping everything in a transaction would slightly change behavior across every run of SQL statements, something worth exploring in a separate effort.

common/common/relations/database_provides.py
```diff
@@ -30,12 +30,25 @@ class _UnsupportedExtraUserRole(status_exception.StatusException):
...
         super().__init__(ops.BlockedStatus(message))
 
 
+class _InvalidDatabaseName(status_exception.StatusException):
+    """Application charm requested an invalid database name"""
+
+    def __init__(self, *, app_name: str, endpoint_name: str, exception_msg: str) -> None:
+        message = (
+            f"{app_name} app requested an invalid database name on "
+            f"{endpoint_name} endpoint: {exception_msg}"
```
carlcsaposs-canonical:
```suggestion
            f"{app_name} app requested invalid database name >20 characters on "
            f"{endpoint_name} endpoint"
```
to keep message short (under 120 characters) so it doesn't get truncated in `juju status` & to give user clear actionable next step (request a shorter database name)

common/common/relations/database_provides.py
```diff
@@ -30,12 +30,25 @@ class _UnsupportedExtraUserRole(status_exception.StatusException):
...
+        )
+        logger.warning(message)
+        super().__init__(ops.BlockedStatus(exception_msg))
```
carlcsaposs-canonical:
```suggestion
        super().__init__(ops.BlockedStatus(message))
```

common/common/relations/database_provides.py
```diff
@@ -79,9 +92,11 @@ def __init__(
         if isinstance(event, ops.RelationBrokenEvent) and event.relation.id == self._id:
             raise _RelationBreaking
         self._database: str = self._databag["database"]
+        self._app_name: str = relation.app.name
+        self._endpoint_name: str = relation.name
         if self._databag.get("extra-user-roles"):
             raise _UnsupportedExtraUserRole(
-                app_name=relation.app.name, endpoint_name=relation.name
+                app_name=self._app_name, endpoint_name=self._endpoint_name
             )
```
carlcsaposs-canonical, line 85: could you raise here if database name > 20 characters so that the _RelationThatRequestedUser object does not get instantiated?

the current pattern/design is _RelationThatRequestedUser is only successfully initialized if the request is valid

common/common/mysql_shell/__init__.py
```diff
@@ -123,17 +127,67 @@ def _get_attributes(self, additional_attributes: dict = None) -> str:
...
         self._run_sql([
-            f"CREATE DATABASE IF NOT EXISTS `{database}`",
             f"CREATE USER `{username}` IDENTIFIED BY '{password}' ATTRIBUTE '{attributes}'",
             f"GRANT ALL PRIVILEGES ON `{database}`.* TO `{username}`",
         ])
-        logger.debug(f"Created {database=} and {username=} with {attributes=}")
+        logger.debug(f"Created {username=} with {attributes=}")
+        return password
+
+    def create_application_database(self, *, database: str, username: str) -> str:
```
carlcsaposs-canonical, line 217:
```suggestion
    def create_application_database_and_user(self, *, database: str, username: str) -> str:
```

common/common/mysql_shell/__init__.py
```diff
@@ -123,17 +127,67 @@ def _get_attributes(self, additional_attributes: dict = None) -> str:
...
+                output_filepath=output_file.relative_to_container,
+            )
+        )
+        with output_file.open("r") as file:
+            rows = json.load(file)
+        output_file.unlink()
+        logger.debug(f"MySQL roles found for {name_pattern=}: {len(rows)}")
+        return {row[0] for row in rows}
+
+    def _create_application_database(self, *, database: str, rolename: str) -> str:
```
carlcsaposs-canonical:
```suggestion
    def _create_application_database(self, *, database: str, rolename: str):
```
nit; return value unused

common/common/mysql_shell/__init__.py
```diff
@@ -123,17 +127,67 @@ def _get_attributes(self, additional_attributes: dict = None) -> str:
...
+        logger.debug(f"Created {username=} with {attributes=}")
+        return password
+
+    def create_application_database(self, *, database: str, username: str) -> str:
+        """Create both the database and the relation user, returning its password."""
+        rolename = f"charmed_dba_{database}"
+        if len(rolename) >= _ROLE_MAX_LENGTH:
+            raise ValueError("Database DBA role longer than 32 characters")
+
+        ________ = self._create_application_database(database=database, rolename=rolename)
```
carlcsaposs-canonical:
```suggestion
        self._create_application_database(database=database, rolename=rolename)
```
nit

common/common/relations/database_provides.py
```diff
@@ -30,12 +30,25 @@ class _UnsupportedExtraUserRole(status_exception.StatusException):
 
     def __init__(self, *, app_name: str, endpoint_name: str) -> None:
         message = (
-            f"{app_name} app requested unsupported extra user role on {endpoint_name} endpoint"
+            f"{app_name} app requested unsupported extra user role on "
+            f"{endpoint_name} endpoint"
```
carlcsaposs-canonical: nit: why wrap? line length is 99 characters https://github.com/canonical/mysql-router-operators/blob/5da2b8bb1d4752430fbd0641a2f5941f1ed5f21d/common/pyproject.toml#L24

common/common/mysql_shell/__init__.py
carlcsaposs-canonical: > Wrapping everything in a transaction

can we do this for every statement we use? I don't know if all of the statements we use support transactions

until then, could be helpful to increase the atomicity in this case

common/common/relations/database_provides.py
sinclert-canonical: I would rather have a distinction between the log message (long format), and the message displayed when doing `juju status` (short format). As it is proposed at the moment:

- Log: _{app} requested an invalid database name on {endpoint} endpoint: DBA role longer than 32 characters_
- Status: _DBA role longer than 32 characters_.

common/common/relations/database_provides.py
sinclert-canonical: Answered in [this comment](https://github.com/canonical/mysql-router-operators/pull/34#discussion_r2325090531).

common/common/relations/database_provides.py
sinclert-canonical: I considered that approach too, to have the exception being raised in the `__init__` method. I am on the fence whether that is a good separation of concerns or not, as it will require the `_RelationThatRequestedUser` class to know implementation details about the `MySQLshell.create_application_database` function.

Maybe exposing a new mysql_shell's `MYSQL_DBA_ROLE_PREFIX` constant with the _charmed\_dba\__ value?

common/common/relations/database_provides.py
sinclert-canonical: Just to make both exceptions have break lines in the same exact places. Easier to see where they differ (i.e. one includes the underlying exception message, while the other do not).

common/common/relations/database_provides.py
paulomach: That's good alternative.

common/common/relations/database_provides.py
paulomach: The constant is a good middle ground, as it will also document reasoning.

paulomach (approved):
Looks good - let's build consensus on open questions, but hold until next week as per discussion

common/common/relations/database_provides.py
sinclert-canonical: Done in commit https://github.com/canonical/mysql-router-operators/pull/34/commits/fb8471416b7888aa2956d94914e5b1662c719b47.

paulomach (approved):
Thank you

sinclert-canonical:
I updated the PR to contained the agreed approach to not limit user database name on +/- 20 characters. In short, we will be adding a two-digit numeric identifier at the end, to avoid collisions with other DBA roles whose first characters coincide.

common/common/mysql_shell/__init__.py
```diff
@@ -123,17 +127,80 @@ def _get_attributes(self, additional_attributes: dict = None) -> str:
...
+        return {row[0] for row in rows}
+
+    def _build_application_database_dba_role(self, database: str) -> str:
+        """Builds the database-level DBA role, given length constraints."""
+        role_prefix = "charmed_dba"
+        role_suffix = "XX"
+
+        role_name_available = _ROLE_MAX_LENGTH - len(role_prefix) - len(role_suffix) - 2
+        role_name_description = database[:role_name_available]
+        role_name_collisions = self._get_mysql_roles(f"{role_prefix}_{role_name_description}_%")
```
carlcsaposs-canonical, line 170: if `create_application_database_and_user` fails part way through or if an uncaught exception is raised later in the same Juju event & the changes aren't saved to the databag, I believe on retry the number will be incremented & a new role will be created

could we use a deterministic hash of the database name instead?

common/common/relations/database_provides.py
carlcsaposs-canonical, line 1: think changes in this file are no longer needed & can be reverted

common/common/mysql_shell/__init__.py
sinclert-canonical: > If an uncaught exception is raised later in the same Juju event & the changes aren't saved to the databag, I believe on retry the number will be incremented & a new role will be created

The databag does not plays a role in this.

I understand the concern about having multiple retries creating ~ similar roles, that is why the role creation is done after the database have been created and not after (just in case such creation fails).

> Could we use a deterministic hash of the database name instead?

The database-level DBA role whole purpose is to allow custom-users, created through a Data Integrator relation, to have admin permissions on a certain database by leveraging the `extra-user-roles` param. For this purpose, the roles must be recognizable, and not just a list of `charmed_dba_<hash>`.

By appending a number at the end of the role name we slightly reduce user's ability to map databases and DBA roles at a glance on a very specific case: when multiple databases coincide on their first set of character. Nevertheless, it it a necessary evil if we want to lift the constraint of 20 character database names to begin with, which is what you complained about to begin with.

In short: we are striking a delicate balance between:
- Role description (`charmed_dba` part).
- Database recognition (`<database>` part).
- Database name de-duplication (`<num>` part).

common/common/relations/database_provides.py
sinclert-canonical: Reverted in commit https://github.com/canonical/mysql-router-operators/pull/34/commits/6e379bbcae028d5ffd267c7c2b8789719515abf5.

common/common/mysql_shell/__init__.py
sinclert-canonical: > If an uncaught exception is raised later in the same Juju event & the changes aren't saved to the databag, I believe on retry the number will be incremented & a new role will be created

The databag does not plays a role in this.

I understand the concern about having multiple retries creating ~ similar roles, but as far as I know, there are no automatic retries when such method fails: retries would need to come from a user manually retrying the `juju relate` command.

To reduce the chance of having legacy DBA roles hanging around, the creation of such roles was proposed to happen after the creation of their assigned databases, reducing the risk of _something going wrong_.

> Could we use a deterministic hash of the database name instead?

The database-level DBA role whole purpose is to allow custom-users, created through a Data Integrator relation, to have admin permissions on a certain database by leveraging the `extra-user-roles` param. For this purpose, the roles must be recognizable, and not just a list of `charmed_dba_<hash>`.

By appending a number at the end of the role name we slightly reduce user's ability to map databases and DBA roles at a glance on a very specific case: when multiple databases coincide on their first set of character. Nevertheless, it it a necessary evil if we want to lift the constraint of 20 character database names to begin with, which is what you complained about to begin with.

In short: we are striking a delicate balance between:
- Role description (`charmed_dba` part).
- Database recognition (`<database>` part).
- Database name de-duplication (`<num>` part).

carlcsaposs-canonical: if an uncaught exception is raised later in the juju event, changes to databags will not be saved

so this method will be called again
https://github.com/canonical/mysql-router-operators/blob/6e379bbcae028d5ffd267c7c2b8789719515abf5/machines/src/relations/deprecated_shared_db_database_provides.py#L297-L301
https://github.com/canonical/mysql-router-operators/blob/6e379bbcae028d5ffd267c7c2b8789719515abf5/machines/src/relations/deprecated_shared_db_database_provides.py#L136-L145

I think we would be safe with a 6-7 character hash—leaving some space for the first part of the unhashed database name

sinclert-canonical: > If an uncaught exception is raised later in the juju event, changes to databags will not be saved

So, you are referring to data that the `create_database_and_user` method will persist on the data-bag, not strictly related to the DBA role proposed in this PR then. If you are afraid of something going wrong when issuing queries to the underlying database, we could wrap the whole thing into a `try-except` block. However, I would argue we should not implement automatic retries.

> I think we would be safe with a 6-7 character hash—leaving some space for the first part of the unhashed database name

That is not the agreement that Mohamed, Mykola, Paulo and I reached.

carlcsaposs-canonical: > data that the `create_database_and_user` method will persist on the data-bag, not strictly related to the DBA role proposed in this PR then

if the data is not persisted on the databag, the user (and therefore the role) will be recreated. Currently, when this happens, the old user is being cleaned up

> That is not the agreement that Mohamed, Mykola, Paulo and I reached.

My understanding from the sync was that we should add a unique identifier but that it was up to us to decide how to implement that identifier

An incrementing identifier based on database state—as opposed to a deterministic identifier—opens up a whole new category of issues (e.g. race conditions if router and server create a user at the same time, extra unused roles being created that don't get cleaned up)

sinclert-canonical: > If the data is not persisted on the databag, the user (and therefore the role) will be recreated. Currently, when this happens, the old user is being cleaned up

I agree, what I am saying is that this concern existed before this PR, and therefore is unrelated to it. Happy to review and approve a follow-up PR that fixes error handling when database or relation-user creation fails, but this is not the place for it.

> My understanding from the sync was that we should add a unique identifier but that it was up to us to decide how to implement that identifier

We agreed to have a short-and-simple numeric identifier at the end (we always talked about 2-3 digits), in order to differentiate between database whose names start by the same set of characters. Part of the agreement was to leave enough place for as many characters as we could fit from the database name, so that the role is "recognizable" by looking at it.

> An incrementing identifier based on database state—as opposed to a deterministic identifier—opens up a whole new category of issues

Paulo and I are chatting about an improvement to greatly reduce the creating of similar DBA roles, whenever a juju relation targeting the same database gets re-related multiple times.

We are planning to check for database existence before creating the DBA role:
- If the database exists: we assume a DBA role must have already been created for it.
- IF the database does not exist: we create both the database and the role.

sinclert-canonical: > If the data is not persisted on the databag, the user (and therefore the role) will be recreated. Currently, when this happens, the old user is being cleaned up

I agree, what I am saying is that this concern existed before this PR, and therefore is unrelated to it. Happy to review and approve a follow-up PR that fixes error handling when database or relation-user creation fails, but this is not the place for it.

> My understanding from the sync was that we should add a unique identifier but that it was up to us to decide how to implement that identifier

We agreed to have a short-and-simple numeric identifier at the end (we always talked about 2-3 digits), in order to differentiate between database whose names start by the same set of characters. Part of the agreement was to leave enough place for as many characters as we could fit from the database name, so that the role is "recognizable" by looking at it.

> An incrementing identifier based on database state—as opposed to a deterministic identifier—opens up a whole new category of issues

Paulo and I are chatting about an improvement to greatly reduce the creating of similar DBA roles, whenever a juju relation targeting the same database gets re-related multiple times.

We are planning to check for database existence before creating the DBA role:
- If the database exists: we assume a DBA role must have already been created for it.
- If the database does not exist: we create both the database and the role.

carlcsaposs-canonical: > what I am saying is that this concern existed before this PR

it did not—before this PR, the user would get cleaned up if the databag changes were not persisted

> database existence before creating the DBA role

this still suffers from a race condition issue between router + server or between multiple router applications. I think it's worth considering sacrificing a few characters of the database name to gain a deterministic identifier (so that we can entirely avoid this category of issues)

sinclert-canonical: > It did not—before this PR, the user would get cleaned up if the databag changes were not persisted

This may require an online chat, because I do not see where this is coming from.

> this still suffers from a race condition issue between router + server or between multiple router applications. I think it's worth considering sacrificing a few characters of the database name to gain a deterministic identifier (so that we can entirely avoid this category of issues)

Yes, and everything suffers from user manually jumping into a MySQL shell and mess around the roles 🤷🏻‍♂️

sinclert-canonical: > We are planning to check for database existence before creating the DBA role.

Just did in commit https://github.com/canonical/mysql-router-operators/pull/34/commits/709f087d6ac58bd2dbd7463798af17aa75f8dd99.

common/common/mysql_shell/__init__.py
sinclert-canonical: Changed in commit https://github.com/canonical/mysql-router-operators/pull/34/commits/709f087d6ac58bd2dbd7463798af17aa75f8dd99.

common/common/mysql_shell/__init__.py
sinclert-canonical: Changed in commit https://github.com/canonical/mysql-router-operators/pull/34/commits/709f087d6ac58bd2dbd7463798af17aa75f8dd99.

common/common/mysql_shell/__init__.py
paulomach: Looks good. Thanks

carlcsaposs-canonical: The race condition still exists—I believe deploying several router applications is a supported use case (e.g. sunbeam)

> may require an online chat

happy to discuss, feel free to add something to my calendar

sinclert-canonical: > The race condition still exists—I believe deploying several router applications is a supported use case (e.g. sunbeam)

AFAIK fetching something to immediately act on it is a common pattern within charms. For example: these same charms fetch user information to delete it right after (see [code](https://github.com/canonical/mysql-router-operators/blob/971cf7dfd95fbd0b43a5b843324eb84d91989f12/common/common/workload.py#L247-L250)). There are race conditions there, as well as in any other place where we follow the pattern of fetch+execute.

I understand your point about race conditions, but unless you provide some other evidence of _how worse this specific race condition is compared to the rest_, I think it is a calculated risk.

> Happy to discuss.

Before doing so:

My understanding from [this comment](https://github.com/canonical/mysql-router-operators/pull/34#discussion_r2354570236) is that uncaught exceptions whenever we issue queries to the MySQL instance could lead to a lack of persistence in the data-bag, which will resort into MySQL Router eventually cleaning up the roles. I get that.

However, I do not think any exception catching is happening on the code prior to this PR. **Where are such exceptions being caught right now, in the current code-base?** Because if they are not, then your concern refers to an error-handling improvement completely unrelated to the code proposed in this PR.

carlcsaposs-canonical: the race conditions I'm referring to are multiple mysql router applications racing each other, or the mysql router app racing the server app—in the case where different databases beginning with the same set of characters are requested at around the same time

given that many deployments deploy several routers, I think the situation where several routers are deployed at the same time (and attempt to create roles at the same time, when the mysql charm first sets endpoints in the databag) is somewhat likely

in the case you linked, only 1 unit is deleting that user—it's not racing with any other unit for the same resource

I think it's worth avoiding the race condition by using a couple more characters of the role name.

Also, what happens if a user requests 100+ databases with the same prefix? I don't think that's super common but I don't think there's a zero probability of that happening

> uncaught exception [...] databag

this is no longer relevant with the addition of the check if the database already exists—although that introduces a new potential issue where the role creation won't be retried if the database creation succeeds but later steps fail

the issue that previously existed was uncaught exceptions (e.g. `dispatch` exits non-zero) causes Juju to not save changes to the databags & retry the Juju event. In those cases, the charm code should handle that situation. The code in `dpe` handles this gracefully by deleting any existing user & retrying

sinclert-canonical: > In the case you linked, only 1 unit is deleting that user—it's not racing with any other unit for the same resource

As you know by now, I was thinking about another type of race conditions: those where there is some change to the MySQL users between fetching and acting on a specific user.

> The race conditions I'm referring to are multiple mysql router applications racing each other, or the mysql router app racing the server app—in the case where different databases beginning with the same set of characters are requested at around the same time

As far as I know, _create database / create user_ operations should be executed in the _primary_, who will replicate these operations to the replicas, so just a single unit should be in charge of executing these operations. Now _what about multiple routers requesting database creation at the ~ same time?_ That would only be a problem if juju events execution is inter-leaved (i.e. `async` operations). Do you know if that's the case?

> This is no longer relevant with the addition of the check if the database already exists.

Thanks for clarifying that this piece of feedback is no longer relevant.

> Although that introduces a new potential issue where the role creation won't be retried if the database creation succeeds but later steps fail

As we already talked in another PR, that is the case for every set of statements issued by the mysql_shell [module](https://github.com/canonical/mysql-router-operators/blob/dpe/common/common/mysql_shell/__init__.py), containing more than one statement (i.e. [these lines](https://github.com/canonical/mysql-router-operators/blob/971cf7dfd95fbd0b43a5b843324eb84d91989f12/common/common/mysql_shell/__init__.py#L132-L134)). _What if some of them succeed and some of them failed?_.

A solutions to this **already existing** issue would be to wrap every execution of SQL statements into a transaction, so we guarantee that either all of them succeed, or all of them failed. However, that is an improvement completely orthogonal to the contents proposed in this PR.

carlcsaposs-canonical: > _database / create user_ operations should be executed in the _primary_

the race is between the call that checks how many roles exist with that database name prefix & creating the database role

"how many roles exist with that database name prefix" is a shared resource, in the sense that multiple Juju applications impact this value

each Juju application acts independently (i.e. the applications can create roles at the same time). So if multiple database names with the same prefix (but a different name) are being created at the same time, the applications could try to use the same integer for the role ID

for example:
mysql router app A checks how many roles exist with prefix `charmed_dba_somelongprefixnam` -> result is 0
mysql router app B checks how many roles exist with prefix `charmed_dba_somelongprefixnam` -> result is 0
mysql router app A creates role `charmed_dba_somelongprefixnam_00` for database `somelongprefixname_foo`
mysql router app B attempts to use role `charmed_dba_somelongprefixnam_00` for database `somelongprefixname_bar`

> already existing issue

regarding user/database creation, this is not an existing issue
if any part of the current `create_application_database_and_user` code (https://github.com/canonical/mysql-router-operators/blob/971cf7dfd95fbd0b43a5b843324eb84d91989f12/common/common/mysql_shell/__init__.py#L126-L137) fails, it will be retried in the next Juju event until all of the statements are successful (and an uncaught exception is not raised later in the juju event)

sinclert-canonical: > The race is between the call that checks how many roles exist with that database name prefix & creating the database role

So we are back to square one, and the type of race condition described in [this comment](https://github.com/canonical/mysql-router-operators/pull/34#discussion_r2358032203): where the users table gets modified between _fetching_ and _acting upon_ user data.

> Each Juju application acts independently (i.e. the applications can create roles at the same time). So if multiple database names with the same prefix (but a different name) are being created at the same time, the applications could try to use the same integer for the role ID

For what you are describing, this is an issue on MySQL backends with multiple MySQL Router applications, where there exist no coordination between them, making them collide. For what I read between the lines, this situation would be different from the one I raised concerns about (i.e. these [lines](https://github.com/canonical/mysql-router-operators/blob/971cf7dfd95fbd0b43a5b843324eb84d91989f12/common/common/workload.py#L247-L250)) as in there is no strict correspondance of *which MySQL Router app* should create the each independent role.

I would argue the way to solve this would be to centralized database creation on the MySQL charm, instead of having it split MySQL Server / MySQL Router. That would also help us reduce the set of permissions we are forced to give the MySQL Router user, avoid any spec deviation (see [spec comment](https://docs.google.com/document/d/1U1tAWx9a7eb0HvdiS7lnSJS8oHqcl6F9PnAoRdcMjmc/edit?disco=AAABoW_7Ajg)).

---

> It will be retried in the next Juju event until all of the statements are successful

That finally answer the question of _where in the current code-base are we catching the exception, and auto-retrying?_ (stated in [this comment](https://github.com/canonical/mysql-router-operators/pull/34#discussion_r2358032203)). The answer is: nowhere, Juju does that implicitly. It is hard to reach any kind of conclusion if questions are not answered.

I would argue the way to solve this is to wrap every set of SQL statements into a transaction, so we do not need to be prepared for partially-apply SQL commands. That would simplify the charm logic, making us propose changes as the ones in this PR.

sinclert-canonical: > I would argue the way to solve this would be to centralized database creation on the MySQL charm

Paulo and I chatted about it yesterday, and decided to address any architectural issues for 8.4

Nevertheless, we considered the situation where multiple MySQL Routers race to create a database on the same underlying MySQL Server, and build exactly the same DBA role name. We addressed those concerns by:

- Moving role creation prior to database creation. This strategy (i.e. creating the potentially conflicting resource first) allow us to know for sure that whenever a database exists, its DBA role must exist too.
- Retrying on the internal database creation method, so that whenever there are race conditions between MySQL Routers, we are still able to deploy + relate without errors.

sinclert-canonical:
PR branch rebased from `dpe` branch, and squashed for clarity.

common/common/mysql_shell/__init__.py
```diff
@@ -123,19 +128,107 @@ def _get_attributes(self, additional_attributes: dict = None) -> str:
...
+    def create_application_database(self, *, database: str, username: str) -> str:
+        """Create both the database and the relation user, returning its password."""
+        for attempt in tenacity.Retrying(
+            retry=tenacity.retry_if_exception_type(ShellDBError),
+            reraise=True,
+            stop=tenacity.stop_after_delay(30),
+            wait=tenacity.wait_fixed(5),
+        ):
+            with attempt:
+                self._create_application_database(database=database)
```
carlcsaposs-canonical: on k8s, we only have 30 seconds for graceful shutdown

imo, if this fails, we should not retry within the hook to avoid blocking graceful shutdown (and potentially not setting the partition when a unit tears down for refresh)

imo would be better to let uncaught exception raise and have juju retry the hook

common/common/mysql_shell/__init__.py
carlcsaposs-canonical: ^

common/common/mysql_shell/templates/get_mysql_databases.py.jinja
```diff
@@ -0,0 +1,10 @@
+import json
+
+result = session.run_sql("SHOW DATABASES")
```
carlcsaposs-canonical, line 3: if there are a lot of databases, will this query be slow?

might be worth using `SHOW DATABASES LIKE 'foo'` to filter by name

common/common/mysql_shell/__init__.py
sinclert-canonical: I am proposing this timeout (30 seconds), and this wait period between retries (5 seconds) because of the [other code](https://github.com/canonical/mysql-router-operators/blob/cd5378cfed63392447d2875035f21c118bec01f3/common/common/workload.py#L493-L499) leveraging `tenacity` I found within the repo. Unless I am missing something, the K8s grateful shutdown concern will also be applicable there.

I am open to remove the explicit retry-logic, and leave Juju implicit hook retries take it from here. However, I would like to know whether there is any difference with the referenced snippet.

common/common/mysql_shell/templates/get_mysql_databases.py.jinja
sinclert-canonical: Not sure doing a `LIKE` pattern matching will speed things up, but open to change it if you show me otherwise.

common/common/mysql_shell/__init__.py
carlcsaposs-canonical: with the other code linked, I think it's a little bit different since it's a wait for condition instead of a retry on failed condition (and afaik there isn't much of an alternative for wait for)—although ideally we wouldn't have that either, since any retry/sleep can reduce/eliminate our time for graceful shutdown

sinclert-canonical: Ok, so in the referenced case _there is no alternative_, while on current one there is, despite both retrying procedure affecting K8s grateful shutdown.

Given that I have little experience writing code that relies on Juju's implicit hook retry, I will take your word for it. If we see those race conditions appearing, we would need to resort to `tenacity`, maybe setting a 20 secs limit.

sinclert-canonical: Done in commit https://github.com/canonical/mysql-router-operators/pull/34/commits/4919c8c4b0fab41732c1f59f91ef06bf7b96ab84.

carlcsaposs-canonical (approved):

common/common/mysql_shell/templates/get_mysql_databases.py.jinja
paulomach: To have a database query be slow to the point we need to worry about in the charm context, we are talking about many thousands of rows in a single column table output. Nothing to worry there folks, let's move on.

sinclert-canonical:
Comment for future us:

It was decided in [this conversation](https://github.com/canonical/mysql-router-operators/pull/34#discussion_r2371676295) to rely on the Juju implicit retry mechanism, instead of the explicit `tenacity.Retrying` decorator, to deal with race concitions in setup where multiple MySQL Routers create databases in a single MySQL Server (given concerns about Kubernetes graceful shutdown).

We should revisit this decision in case we see any issues with Juju's implicit retries (i.e. they change in Juju 4, they are limited to a very low number, etc).

PR #40, opened 2025-09-30
[DPE-7796] Define Terraform modules

sinclert-canonical:
This PR defines the Terraform modules of both MySQL Router VM & K8s substrates, following the[ CC006 specification](https://docs.google.com/document/d/1EG71A2pJ244PQRaGVzGj7Mx2B_bzE4U_OSqx4eeVI1E/edit?tab=t.0), which outlines how to organize Terrraform variables and outputs at the _charm_ level.

From the list of optional input variables:
- [x] base
- [ ] expose
- [ ] resources
- [ ] machines
- [ ] endpoint_bindings 
- [ ] storage

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -36,6 +36,8 @@ jobs:
         uses: actions/checkout@v5
       - name: Set up Terraform CLI
         uses: hashicorp/setup-terraform@v3
+        with:
+          terraform_wrapper: false
       - name: Install tox & poetry
         run: |
           pipx install tox
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -34,13 +34,17 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v5
+      - name: Set up Terraform CLI
+        uses: hashicorp/setup-terraform@v3
       - name: Install tox & poetry
         run: |
           pipx install tox
           pipx install poetry
       - name: Run linters
         working-directory: ${{ matrix.path }}
-        run: tox run -e lint
+        run: |
+          tox run -e lint
+          tox run -e lint-terraform
 
   unit-test:
     strategy:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -34,10 +34,9 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v5
-      - name: Set up Terraform CLI
-        uses: hashicorp/setup-terraform@v3
-        with:
-          terraform_wrapper: false
+      - name: Install terraform
+        run: |
+          sudo snap install terraform --classic
       - name: Install tox & poetry
         run: |
           pipx install tox
--- a/kubernetes/terraform/variables.tf
+++ b/kubernetes/terraform/variables.tf
@@ -1,0 +1,46 @@
+variable "model_name" {
+  description = "Name of the juju model to deploy to"
+  type        = string
+}
+
+variable "app_name" {
+  description = "Name of the juju application"
+  type        = string
+  default     = "mysql-router"
+}
+
+variable "base" {
+  description = "Application base"
+  type        = string
+  default     = "ubuntu@22.04"
+}
+
+variable "config" {
+  description = "Application configuration. Details at https://charmhub.io/mysql-router-k8s/configurations"
+  type        = map(string)
+  default     = {}
+}
+
+variable "constraints" {
+  description = "Juju constraints for the application"
+  type        = string
+  default     = "arch=amd64"
+}
+
+variable "channel" {
+  description = "Charm channel to deploy from"
+  type        = string
+  default     = "8.0/stable"
+}
+
+variable "revision" {
+  description = "Charm revision to deploy"
+  type        = number
+  default     = null
+}
+
+variable "units" {
+  description = "Number of units to deploy"
+  type        = number
+  default     = 1
+}
--- a/kubernetes/terraform/variables.tf
+++ b/kubernetes/terraform/variables.tf
@@ -6,7 +6,7 @@ variable "model_name" {
 variable "app_name" {
   description = "Name of the juju application"
   type        = string
-  default     = "mysql-router"
+  default     = "mysql-router-k8s"
 }
 
 variable "base" {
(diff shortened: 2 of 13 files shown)


.github/workflows/ci.yaml
```diff
@@ -34,13 +34,19 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v5
+      - name: Set up Terraform CLI
+        uses: hashicorp/setup-terraform@v3
+        with:
+          terraform_wrapper: false
```
carlcsaposs-canonical: should we install terraform from snap instead, to reduce security risk of third-party action?

`hashicorp` is probably trustworthy, but don't think an action is necessary here—and a third-party action increases the surface area for an attack in our workflows (even outside of this job/workflow) significantly

kubernetes/terraform/variables.tf
```diff
@@ -0,0 +1,46 @@
+variable "model_name" {
+  description = "Name of the juju model to deploy to"
+  type        = string
+}
+
+variable "app_name" {
+  description = "Name of the juju application"
+  type        = string
+  default     = "mysql-router"
```
carlcsaposs-canonical:
```suggestion
  default     = "mysql-router-k8s"
```
to match default with `juju deploy`

machines/terraform/variables.tf
```diff
@@ -0,0 +1,46 @@
...
+variable "constraints" {
+  description = "Juju constraints for the application"
+  type        = string
+  default     = "arch=amd64"
+}
+
+variable "channel" {
+  description = "Charm channel to deploy from"
+  type        = string
+  default     = "dpe/edge"
```
carlcsaposs-canonical:
```suggestion
  default     = "dpe/candidate"
```

kubernetes/terraform/variables.tf
```diff
@@ -0,0 +1,46 @@
...
+variable "config" {
+  description = "Application configuration. Details at https://charmhub.io/mysql-router-k8s/configurations"
+  type        = map(string)
+  default     = {}
+}
+
+variable "constraints" {
+  description = "Juju constraints for the application"
+  type        = string
+  default     = "arch=amd64"
```
carlcsaposs-canonical, line 27: question: does this have any effect? I thought arch constraints needed to be applied on the model instead of on the app, but not sure

carlcsaposs-canonical (approved):

.github/workflows/ci.yaml
sinclert-canonical: Totally fine by me, despite the action being more popular within Canonical repos (see [action](https://github.com/search?q=org%3Acanonical+language%3AYAML+%22hashicorp%2Fsetup-terraform%22&type=code) vs [snap install](https://github.com/search?q=org%3Acanonical+language%3AYAML+%22sudo+snap+install+--classic+terraform%22&type=code)). Given that I will define the same linters steps in all MySQL repos, we should have quorum about this.

@paulomach opinion?

kubernetes/terraform/variables.tf
sinclert-canonical: IIRC constraints, including architecture ones, can be specified at:
- A) Model level, via `set-model-constraints` command.
- B) Application level, via `--constraints` argument.

The referenced [spec](https://docs.google.com/document/d/1EG71A2pJ244PQRaGVzGj7Mx2B_bzE4U_OSqx4eeVI1E/edit?tab=t.0) includes application level constraints as a mandatory variable, so we should preserve it. Are you worried about the default value proposed?

.github/workflows/ci.yaml
carlcsaposs-canonical: fyi it looks like the snap is more popular (https://github.com/search?q=org%3Acanonical+language%3AYAML+sudo+snap+install+terraform&type=code)

also looks like some people are using `opentofu`, maybe we should consider that

paulomach: lets use snap as this is the documented method of using tf in ubuntu.

kubernetes/terraform/versions.tf
```diff
@@ -0,0 +1,10 @@
+terraform {
+  required_version = ">= 1.6.6"
+
+  required_providers {
+    juju = {
+      source  = "juju/juju"
+      version = ">= 0.20.0"
```
paulomach: Just as a heads up, juju terraform provider is at in `v0.23` but `v1.0.0` is scheduled to be released this month with [breaking changes](https://discourse.charmhub.io/t/upcoming-changes-in-the-juju-terraform-provider/18232/1), so we may need to update this somewhat soon.

machines/terraform/versions.tf
```diff
@@ -0,0 +1,10 @@
+terraform {
+  required_version = ">= 1.6.6"
+
+  required_providers {
+    juju = {
+      source  = "juju/juju"
+      version = ">= 0.20.0"
```
paulomach: Same as [k8s](https://github.com/canonical/mysql-router-operators/pull/40/files#r2396441357)

paulomach (approved):
Nice. Let's keep an eye on juju provider version bump and make sure we are not impacted by breaking changes

kubernetes/terraform/versions.tf
sinclert-canonical: Good point!

For what I can tell, the only breaking change that will affect us is the usage of model UUIDs instead of model names. Will make sure to suffix every provider version with a `< 1.0.0` selector, to avoid future problems.

.github/workflows/ci.yaml
sinclert-canonical: Done in commit https://github.com/canonical/mysql-router-operators/pull/40/commits/80b0664ba54edeb96766f72863ffdf6c0d31dc67.

machines/terraform/variables.tf
sinclert-canonical: Done in commit https://github.com/canonical/mysql-router-operators/pull/40/commits/80b0664ba54edeb96766f72863ffdf6c0d31dc67.

kubernetes/terraform/variables.tf
sinclert-canonical: Done in commit https://github.com/canonical/mysql-router-operators/pull/40/commits/80b0664ba54edeb96766f72863ffdf6c0d31dc67.

kubernetes/terraform/versions.tf
sinclert-canonical: Done in commit https://github.com/canonical/mysql-router-operators/pull/40/commits/bc83fc844f19ac05aa3d9e8cc8e6677a2df50cc2.

PR #53, opened 2025-10-24
Use python 3.10 syntax

carlcsaposs-canonical:


common/common/abstract_charm.py
```diff
@@ -99,7 +99,7 @@ def __init__(self, *args) -> None:
 
     @property
     @abc.abstractmethod
-    def _subordinate_relation_endpoint_names(self) -> typing.Optional[typing.Iterable[str]]:
+    def _subordinate_relation_endpoint_names(self) -> collections.abc.Collection[str] | None:
```
sinclert-canonical, line 102: Was `Iterable` not precised enough?

sinclert-canonical (approved):

common/common/abstract_charm.py
carlcsaposs-canonical: Iterable was incorrect in lifecycle.py, since we were testing `in` multiple times

(so generator would pass type check but fail at runtime)

lifecycle.py needs `__contains__` and `__bool__` (implicit from `__len__`); Collection is the most abstract class that guarantees that

paulomach (approved):

PR #55, opened 2025-10-27
Update ops to 2.10

carlcsaposs-canonical:
Will update ops further in a separate PR (to make diff easier to review, since changes to unit tests will be needed with ops-scenario update)


carlcsaposs-canonical:
Think it might be worth waiting until we have a new stable release before merging

This is a pretty significant change & would like to avoid multiple significant changes (refresh v3) in a single stable release in case there are any issues & we want to isolate part of the changes

carlcsaposs-canonical:
Looks like ops-scenario has issues with `relation.active` (returning True when it should be returning False); investigating further

astrojuanlu:
Closing in favor of #212.

PR #75, opened 2025-12-11
[DPE-9072] allow rebootstrap on metadata removal

paulomach:
## Issue

See issue #74

By design mysql-router is not allowing rebootstrap if router is removed from cluster.
While this is by design, is undocumented and frustrate user expectations to have a charm that just works.

## Solution

Use detection for router metadata to ensure rebootstrap happens.

### DIFF
--- a/common/common/workload.py
+++ b/common/common/workload.py
@@ -405,7 +405,7 @@ class RunningWorkload(Workload):
         socket_file_exists = self._container.path("/run/mysqlrouter/mysql.sock").exists()
         require_rebootstrap = is_charm_exposed == socket_file_exists
         # If the router is not in the cluster set, disable to restart it
-        # This can happen when the server is scalled to zero and back
+        # This can happen when the server is scaled to zero and back
         require_rebootstrap = require_rebootstrap or not self.shell.is_router_in_cluster_set(
             self._router_id
         )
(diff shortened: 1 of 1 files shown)


sinclert-canonical (approved):
My understanding is that this fix targets a very specific environment: Kubernetes, where application units can be scaled down to zero, without removing the application completely. Correct?

paulomach:
> My understanding is that this fix targets a very specific environment: Kubernetes, where application units can be scaled down to zero, without removing the application completely. Correct?

Correct, but not only. In this particular case, when the server scale down to zero, it dissolves the cluster, so router metadata is lost. This is deliberate so it can accommodate changes in the config, like cluster name change. When the server is scale up, it recreates the cluster, picking up any new config. 

But the change in behavior proposed will affect other cases, where for whatever reason the router won't be in cluster metadata (e.g. manual removal), which is not k8s exclusive.

This can be further simplified to make the router rebootstrap always, regardless of error origin, so it just works. But that's a big refactor and we don't have budget right now.

common/common/workload.py
```diff
@@ -404,6 +404,11 @@ def reconcile(
         is_charm_exposed = self._charm.is_externally_accessible(event=event)
         socket_file_exists = self._container.path("/run/mysqlrouter/mysql.sock").exists()
         require_rebootstrap = is_charm_exposed == socket_file_exists
+        # If the router is not in the cluster set, disable to restart it
+        # This can happen when the server is scalled to zero and back
```
astrojuanlu:
```suggestion
        # This can happen when the server is scaled to zero and back
```

astrojuanlu (approved):

PR #76, opened 2025-12-15
[DPE-8877] Upgrade MySQL to 8.0.44

astrojuanlu:
Mirrors https://github.com/canonical/mysql-k8s-operator/pull/706, https://github.com/canonical/mysql-operator/pull/728

### DIFF
--- a/kubernetes/refresh_versions.toml
+++ b/kubernetes/refresh_versions.toml
@@ -1,4 +1,4 @@
 # https://canonical-charm-refresh.readthedocs-hosted.com/latest/refresh-versions-toml/
 
 charm_major = 1
-workload = "8.0.42"
+workload = "8.0.44"
--- a/kubernetes/tests/integration/test_upgrade.py
+++ b/kubernetes/tests/integration/test_upgrade.py
@@ -120,12 +120,16 @@ async def test_upgrade_from_edge(ops_test: OpsTest, charm) -> None:
 
     # Refresh will be incompatible on PR CI (not edge CI) since unreleased charm versions are
     # always marked as incompatible
-    if (
-        refresh_order[0].workload_status == "blocked"
-        and "incompatible" in refresh_order[0].workload_status_message
+    if refresh_order[0].workload_status == "blocked" and (
+        "incompatible" in refresh_order[0].workload_status_message
+        or "missing/incorrect OCI resource" in refresh_order[0].workload_status_message
     ):
         logger.info("Running force-refresh-start action with check-compatibility=false")
-        await run_action(refresh_order[0], "force-refresh-start", **{"check-compatibility": False})
+        await run_action(
+            refresh_order[0],
+            "force-refresh-start",
+            **{"check-compatibility": False, "check-workload-container": False},
+        )
 
     logger.info("Wait for first unit to upgrade")
     await ops_test.model.wait_for_idle([MYSQL_ROUTER_APP_NAME], idle_period=30, timeout=TIMEOUT)
@@ -184,9 +188,10 @@ async def test_fail_and_rollback(ops_test: OpsTest, charm, continuous_writes) ->
     await ops_test.model.block_until(
         lambda: refresh_order[0].workload_status == "blocked", timeout=TIMEOUT
     )
-    assert "incompatible" in refresh_order[0].workload_status_message, (
-        "mysql router application status not indicating that refresh incompatible"
-    )
+    assert (
+        "incompatible" in refresh_order[0].workload_status_message
+        or "missing/incorrect OCI resource" in refresh_order[0].workload_status_message
+    ), "mysql router application status not indicating that refresh incompatible"
 
     logger.info("Ensure continuous writes while in failure state")
     await ensure_all_units_continuous_writes_incrementing(ops_test)
@@ -242,6 +247,8 @@ def create_invalid_upgrade_charm(charm_file: str | pathlib.Path) -> None:
         with zipfile.Path(charm_zip, "refresh_versions.toml").open("rb") as file:
             versions = tomli.load(file)
 
+    # Format is not yet documented,
+    # see https://canonical-charm-refresh.readthedocs-hosted.com/latest/refresh-versions-toml/
     versions["charm"] = "8.0/0.0.0"
 
     with zipfile.ZipFile(charm_file, mode="a") as charm_zip:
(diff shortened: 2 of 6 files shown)


astrojuanlu:
Current status:

```
blocked: `juju refresh` was run with missing/incorrect OCI resource.
```

kubernetes/refresh_versions.toml
```diff
@@ -1,4 +1,4 @@
 # https://canonical-charm-refresh.readthedocs-hosted.com/latest/refresh-versions-toml/
 
 charm_major = 1
-workload = "8.0.42"
+workload = "8.0.44"
```
astrojuanlu, line 4: Even without this change, the tests were passing locally 🤔

astrojuanlu: Nonsense. I was running the tests locally with an old charm.

paulomach:
@astrojuanlu , as part of this upgrade, [this bug workaround](https://github.com/canonical/mysql-router-operators/blob/0e8abaf9b338a84f53b3b3de3df570761108ecc4/common/common/workload.py#L275-L281) can be removed, as it was fixed in 8.0.43+

kubernetes/tests/integration/test_upgrade.py
```diff
@@ -120,12 +120,16 @@ async def test_upgrade_from_edge(ops_test: OpsTest, charm) -> None:
 
     # Refresh will be incompatible on PR CI (not edge CI) since unreleased charm versions are
     # always marked as incompatible
-    if (
-        refresh_order[0].workload_status == "blocked"
-        and "incompatible" in refresh_order[0].workload_status_message
+    if refresh_order[0].workload_status == "blocked" and (
+        "incompatible" in refresh_order[0].workload_status_message
+        or "missing/incorrect OCI resource" in refresh_order[0].workload_status_message
```
astrojuanlu: I had to add this part, otherwise the test wouldn't pass.

It wasn't very straightforward for me to understand https://github.com/canonical/charm-refresh/blob/f19e5189bb4a8bd0a32a3a764e7d394f1273b27f/charm_refresh/_main.py#L979-L999 but from what I could see, this was the first time this code path was actually exercised? But would love to have further guidance.

kubernetes/tests/integration/test_upgrade.py
```diff
@@ -120,12 +120,16 @@ async def test_upgrade_from_edge(ops_test: OpsTest, charm) -> None:
...
     ):
         logger.info("Running force-refresh-start action with check-compatibility=false")
-        await run_action(refresh_order[0], "force-refresh-start", **{"check-compatibility": False})
+        await run_action(
+            refresh_order[0],
+            "force-refresh-start",
+            **{"check-compatibility": False, "check-workload-container": False},
```
astrojuanlu: Connected to my comment above.

kubernetes/tests/integration/test_upgrade.py
```diff
@@ -184,9 +188,10 @@ async def test_fail_and_rollback(ops_test: OpsTest, charm, continuous_writes) ->
     await ops_test.model.block_until(
         lambda: refresh_order[0].workload_status == "blocked", timeout=TIMEOUT
     )
-    assert "incompatible" in refresh_order[0].workload_status_message, (
-        "mysql router application status not indicating that refresh incompatible"
-    )
+    assert (
+        "incompatible" in refresh_order[0].workload_status_message
+        or "missing/incorrect OCI resource" in refresh_order[0].workload_status_message
+    ), "mysql router application status not indicating that refresh incompatible"
```
astrojuanlu: Interestingly, after creating the incompatible charm, the error message was still "missing/incorrect OCI resource".

kubernetes/tests/integration/test_upgrade.py
```diff
@@ -242,6 +247,8 @@ def create_invalid_upgrade_charm(charm_file: str | pathlib.Path) -> None:
         with zipfile.Path(charm_zip, "refresh_versions.toml").open("rb") as file:
             versions = tomli.load(file)
 
+    # Format is not yet documented,
+    # see https://canonical-charm-refresh.readthedocs-hosted.com/latest/refresh-versions-toml/
     versions["charm"] = "8.0/0.0.0"
```
astrojuanlu: Wanted to understand what the `charm` key should be, but the docs are incomplete, and there's currently no such key in `refresh_versions.toml`

astrojuanlu (commented):
Could use some help with test_upgrade.py before proceeding

kubernetes/tests/integration/test_upgrade.py
carlcsaposs-canonical: "missing/incorrect OCI resource" is not an expected status in the integration tests—I think there's a bug somewhere (in charm-refresh, juju, jubilant/pytest-operator, the charm, or the test code) that should be fixed

more context: https://canonical-charm-refresh.readthedocs-hosted.com/latest/user-experience/status/#unknown-workload

carlcsaposs-canonical: I think the cause is `resources` missing from `.refresh` in the tests
https://github.com/canonical/mysql-router-operators/blob/8427a1f1ff020c1f53f2300e48854087951977fe/kubernetes/tests/integration/test_upgrade.py#L101

example of `resources`: https://github.com/canonical/postgresql-k8s-operator/blob/bfb6b691259f17db58180822995f4305deebb86f/tests/integration/ha_tests/test_upgrade.py#L110

kubernetes/tests/integration/test_upgrade.py
carlcsaposs-canonical: here's some info in the meantime until the docs are added: https://docs.google.com/document/d/1Jv1jhWLl8ejK3iJn7Q3VbCIM9GIhp8926bgXpdtx-Sg/edit?pli=1&tab=t.0

in case you're curious, `charm` is added during pack time using the charm refresh compatibility version stored as a git tag (so `charm` can be set after the release [when the version is determined] but before pack time)

kubernetes/tests/integration/test_upgrade.py
astrojuanlu: > "missing/incorrect OCI resource" is not an expected status in the integration tests

Understood, thank you!

> I think the cause is resources missing from .refresh in the tests

will have a look

kubernetes/tests/integration/test_upgrade.py
astrojuanlu: Added that explanation

astrojuanlu:
We should be good now!

sinclert-canonical (approved):

paulomach (approved):
Thanks!

PR #77, opened 2025-12-17
[MISC] Adopt packaging standards

astrojuanlu:
Mirrors https://github.com/canonical/mysql-k8s-operator/pull/680/ and https://github.com/canonical/mysql-operator/pull/717/, but leaves the `[tool.poetry.dependencies]` so that charmcraftlocal can work properly.


kubernetes/pyproject.toml
```diff
@@ -1,6 +1,57 @@
...
+name = "mysql-router-k8s-operator"
+version = "0.0.1"
+dynamic = ["dependencies", "requires-python"]
+
+[dependency-groups]
+charm-libs = [
+    # data_platform_libs/v0/data_interfaces.py
+    "ops>=2.0.0",
+    # grafana_k8s/grafana_dashboard.py
+    "cosl>= 0.0.50",
```
sinclert-canonical: Nit:
```diff
-    "cosl>= 0.0.50",
+    "cosl>=0.0.50",
```

kubernetes/pyproject.toml
```diff
@@ -1,6 +1,57 @@
...
+    # tempo_coordinator_k8s/v0/charm_tracing.py
+    "opentelemetry-exporter-otlp-proto-http==1.21.0",
+    # tempo_coordinator_k8s/v0/tracing.py
+    "pydantic",
+    # tls_certificates_interface/v1/tls_certificates.py
+    "cryptography",
+    "jsonschema",
+]
+format = [
+    "ruff~=0.12.11",
```
sinclert-canonical, line 24: I think this one should be reduced to just `0.12` if we are using the ~= specifier.

kubernetes/pyproject.toml
```diff
@@ -1,6 +1,57 @@
...
+]
+lint = [
+    "ruff~=0.12.11",
```
sinclert-canonical, line 27: I think this one should be reduced to just `0.12` if we are using the ~= specifier.

kubernetes/pyproject.toml
```diff
@@ -1,6 +1,57 @@
...
+    "pytest~=8.4",
+    "pytest-xdist~=3.8",
+    "pytest-forked~=1.6",  # Workaround to deal with memory leaks in scenario tests
+    "pytest-cov~=6.2",
+    "ops-scenario==6.0.3",  # 6.0.4 requires ops >= 2.12
+    "ops~=2.9",
+]
+integration = [
+    "pytest~=8.4",
+    "pytest-operator~=0.43.1",
```
sinclert-canonical, line 40: I think this one should be reduced to just `0.43` if we are using the ~= specifier.

kubernetes/pyproject.toml
```diff
@@ -1,6 +1,57 @@
...
+    "requests~=2.32",
+    "juju~=3.6",
+    "mysql-connector-python~=9.5",
+    "pyyaml~=6.0",
+    "tenacity~=9.1",
+    "allure-pytest~=2.15",
+    "allure-pytest-default-results~=0.1.3",
```
sinclert-canonical, line 47: I think this one should be reduced to just `0.1` if we are using the ~= specifier.

kubernetes/pyproject.toml
```diff
@@ -1,6 +1,57 @@
...
+    "tomli~=2.2",
+    "tomli-w~=1.2",
+]
+build-refresh-version = [
+    "charm-refresh-build-version~=0.4.0",
```
sinclert-canonical, line 52: I think this one should be reduced to just `0.4` if we are using the ~= specifier.

machines/pyproject.toml
sinclert-canonical, line 1: Exact same comments about the Kubernetes equivalent, when it comes to versions using ~=.

kubernetes/pyproject.toml
astrojuanlu: I was preserving the original bounds (even in cases where the lower bound is more flexible, as discussed in https://github.com/canonical/mysql-k8s-operator/pull/680#discussion_r2548887925.

The ones you highlighted would loosen up the upper bound too. Should I proceed?

sinclert-canonical: OK I see you left the patch version on the MySQL VM / K8s sibling PRs. Let's go with your approach.

sinclert-canonical (approved):

paulomach (approved):

PR #84, opened 2026-02-10
[MISC] Bump mysql-connector-python to 9.5

sinclert-canonical:
This PR bumps `mysql-connector-python` to version 9.5 in the VM operator. This dependency is the one holding back others that needs to be updated for the build process to be fixed.

The error is only visible when building from source (i.e. using charmcraft, and not chamcraftcache):

```log
ModuleNotFoundError: No module named 'pkg_resources'
```

---

Alternative to https://github.com/canonical/mysql-router-operators/pull/83.


astrojuanlu (approved):
💯 

This probably comes because setuptools recently removed pkg_resources https://github.com/pypa/setuptools/blob/v82.0.0/NEWS.rst#v8200

paulomach (approved):

PR #86, opened 2026-02-11
[DPE-8811] Branch out MySQL Router 8.4

sinclert-canonical:
This PR formally branches out MySQL Router 8.4 codebase from 8.0 one, by removing all references to the latter.

- Bumping the MySQL binaries to 8.4.
- Bumping the build platform to 24.04.
- Bumping the required Python to 3.12+.

Depends on:

- https://github.com/canonical/mysql-router-operators/pull/82.
- https://github.com/canonical/mysql-test-app/pull/54.

PR #87, opened 2026-02-11
[MISC] Fix _Path wrapper to work on Python 3.12

astrojuanlu:
The root cause is that in Python 3.12 `PurePath.__init__()` uses lazy initialization for performance, and it only stores `_raw_paths` (a list of unjoined string paths) during `__init__`. The custom `__new__` method was interfering with this new logic.

Managed to rework the wrapper to avoid using `__new__` altogether and hopefully brought some clarity. But pathlib doesn't support subclassing in Python <= 3.11, so this code only works on 3.12 onwards.


sinclert-canonical:
Proposed alternative approach in PR https://github.com/canonical/mysql-router-operators/pull/86.

Apologies for sharing the burden 😅

PR #89, opened 2026-02-13
[MISC] Refactor K8s tests to use `series`

sinclert-canonical:
This PR refactors the K8s integration tests to use `series` instead of the `base` argument. I do not exactly know why, but the version of `ops` that we are using (2.9.0) behaves differently whether one or the other argument is specified. When `base` is used, it seems that the library is not able to detect multi-platform charms

Additionally, I cleaned up the `CHARM_UBUNTU_BASE` env. variable from the VM tests.

---

Unblocks https://github.com/canonical/mysql-router-operators/pull/88.

### DIFF
--- a/machines/spread.yaml
+++ b/machines/spread.yaml
@@ -86,7 +86,6 @@ backends:
     # Manually pass specific environment variables
     environment:
       CI: '$(HOST: echo $CI)'
-      CHARM_UBUNTU_BASE: '$(HOST: echo $CHARM_UBUNTU_BASE)'
       UBUNTU_PRO_TOKEN: '$(HOST: echo $UBUNTU_PRO_TOKEN)'
       LANDSCAPE_ACCOUNT_NAME: '$(HOST: echo $LANDSCAPE_ACCOUNT_NAME)'
       LANDSCAPE_REGISTRATION_KEY: '$(HOST: echo $LANDSCAPE_REGISTRATION_KEY)'
@@ -109,7 +108,6 @@ environment:
   PATH: $PATH:$(pipx environment --value PIPX_BIN_DIR)
   CONCIERGE_JUJU_CHANNEL/juju36: 3.6/stable
   CONCIERGE_JUJU_CHANNEL/juju29: 2.9/stable
-  CHARM_UBUNTU_BASE/juju36,juju29: 22.04
 prepare: |
   snap refresh --hold
   chown -R root:root "$SPREAD_PATH"
--- a/kubernetes/tests/integration/conftest.py
+++ b/kubernetes/tests/integration/conftest.py
@@ -13,11 +13,21 @@ logger = logging.getLogger(__name__)
 
 
 @pytest.fixture
-def charm():
+def ubuntu_base():
+    return "ubuntu@22.04"
+
+
+@pytest.fixture
+def series():
+    return "jammy"
+
+
+@pytest.fixture
+def charm(ubuntu_base):
     # Return str instead of pathlib.Path since python-libjuju's model.deploy(), juju deploy, and
     # juju bundle files expect local charms to begin with `./` or `/` to distinguish them from
     # Charmhub charms.
-    return f"./mysql-router-k8s_ubuntu@22.04-{architecture.architecture}.charm"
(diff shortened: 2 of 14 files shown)


astrojuanlu (approved):
LGTM

taurus-forever (approved):
LGTM to unblock it, but plaease sync with @carlcsaposs-canonical here.
It tings me bells but I do not recall details here.

JFYI: we should use base everywhere as a modern and proper way. Series should R.I.P.
If base are broken -> bugreport it. I see no changes in libjuju 2.9 since [Dec 2024](https://pypi.org/project/juju/#history) 

Q: why don't you migrate to modern libjuju 3+ ?

machines/spread.yaml
```diff
@@ -109,7 +108,6 @@ environment:
   PATH: $PATH:$(pipx environment --value PIPX_BIN_DIR)
   CONCIERGE_JUJU_CHANNEL/juju36: 3.6/stable
   CONCIERGE_JUJU_CHANNEL/juju29: 2.9/stable
-  CHARM_UBUNTU_BASE/juju36,juju29: 22.04
```
taurus-forever, line 112: Hm... curious how it works now.

sinclert-canonical: That was a leftover from the time when we ditched the Ubuntu 20.04 base, due to a Python bug.

Until now, we had a way to inject the Ubuntu base, via env. variable. However, that is not used anymore, as we are hard-coding Ubuntu 22.04 base. I proposed to get rid of that, and keep VM & K8s `conftest.py` in-sync.

sinclert-canonical:
> JFYI: we should use base everywhere as a modern and proper way. Series should R.I.P.
> If base are broken -> bugreport it. I see no changes in libjuju 2.9 since [Dec 2024](https://pypi.org/project/juju/#history)

If it was by me, we would be using _base_ everywhere. However, due to a combination of old `ops` 2.9.0 library, and a new multi-platform MySQL Test App, using _base_ crash constantly (see [blocked PR](https://github.com/canonical/mysql-router-operators/pull/88) errors).

I took a look at how you guys are dealing with the multi-platform "Test app" situation in PGBouncer, and turns out you are skipping both _series_ and _base_ (see [code](https://github.com/canonical/pgbouncer-operator/blob/4f9cb516d697199e5310a790ecb2a0b503fbec39/tests/integration/test_charm.py#L36-L38)). Not sure whether that would work with the version of `ops` we have, but given that the VM substrate already leverages _series_ and works.... I figured we should have the same in both.

> Q: why don't you migrate to modern libjuju 3+ ?

Please, correct me if I am wrong, but I assume you are referring to `ops`, and not `juju` library. To the best of my understanding, the problem lies in the dependency between `ops` and `ops-scenario`, and it is stated [here](https://github.com/canonical/mysql-router-operators/blob/b57ff202d2eaca0a676c60b973cb81b2b9096d1c/machines/pyproject.toml#L34).

taurus-forever:
> Please, correct me if I am wrong, but I assume you are referring to `ops`, and not `juju` library. 

I want to see both on version 3. :-D

* the latest pylijuju is 3.6.1.3: https://pypi.org/project/juju/#history
* Ops 3.5.2: https://github.com/canonical/operator/releases

@dragomirp managed to migrate entire PG stack to [Ops 3](https://github.com/canonical/postgresql-operator/blob/main/pyproject.toml#L11) (even subordinated PgBouncer isolating Focal usecase...).
Also AFAIK we are using [pylijuju 3](https://github.com/canonical/postgresql-operator/blob/main/pyproject.toml#L70) there as well.

Did I miss something?

sinclert-canonical:
> Did I miss something?

Migrating to ops 3 (or even the latest version of ops 2) in this repository will be time-consuming and painful, due to how we are orchestrating the unit-tests and the +/- 1000 combinations in VM & +/- 8000 combinations in K8s. I dedicated an afternoon to see what would take, and the result was not pretty.

Regarding the task itself: we have a [JIRA ticket](https://warthogs.atlassian.net/browse/DPE-8714) about it, which was originally assigned to Carl, but recently assigned to me (see ticket history). However, I can already tell you that, to me, it has the **lowest priority of all my tickets**. Releasing MySQL Server 8.4 + MySQL Router 8.4 is what I am laser focus on.

PR #103, opened 2026-03-13
[MISC] Set trust attribute to true in TF configs

hk21702:
## Issue

The existing TF configs currently don't work since the charms need to be deployed with trust.

For example, on K8s, we get the following `juju status` output: 
```
mysql-router                   blocked      3  mysql-router-k8s      8.4/edge       878  10.86.228.99   no       Run `juju trust mysql-router --scope=cluster`. Needed for in-place refreshes
```

## Solution

Set the `trust` attribute to true in both Terraform config files for the `mysql-router` application.

### DIFF
--- a/kubernetes/terraform/main.tf
+++ b/kubernetes/terraform/main.tf
@@ -1,6 +1,7 @@
 resource "juju_application" "mysql_router" {
   model_uuid = var.model
   name       = var.app_name
+  trust      = true
 
   charm {
     name     = "mysql-router-k8s"


sinclert-canonical (approved):

paulomach (changes requested):
Thank you @hk21702 - iirc `trust` is just need for k8s models

hk21702:
Woops, fixed!

paulomach (approved):

sinclert-canonical:
Tests are very flaky nowadays, but this is safe to merge.

PR #122, opened 2026-03-31
[MISC] 8.0 - Downgrade CI runners to Ubuntu 22.04

sinclert-canonical:
This PR follows the lead of MySQL Server PR https://github.com/canonical/mysql-operators/pull/198 in order to temporarily downgrade CI runners to Ubuntu 22.04, until the AppArmor fix get backported to Ubuntu 24.04 (probably after 26.04 gets released).

### DIFF
--- a/kubernetes/spread.yaml
+++ b/kubernetes/spread.yaml
@@ -80,16 +80,20 @@ backends:
       EOF
 
       sudo passwd --delete "$USER"
+      
+      sudo systemctl reload ssh || sudo systemctl restart ssh
 
       ADDRESS localhost
     # HACK: spread does not pass environment variables set on runner
     # Manually pass specific environment variables
     environment:
       CI: '$(HOST: echo $CI)'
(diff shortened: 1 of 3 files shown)


paulomach (approved):

kubernetes/spread.yaml
```diff
@@ -129,4 +135,5 @@ prepare-each: |
   # Unable to set constraint on all models because of Juju bug:
   # https://bugs.launchpad.net/juju/+bug/2065050
   juju set-model-constraints arch="$(dpkg --print-architecture)"
-# Only restore on lxd backend—no need to restore on CI
```
astrojuanlu, line 132: I don't quite understand what this comment is about, but noting it in case you want to retain it

astrojuanlu (approved):

kubernetes/spread.yaml
sinclert-canonical: I do not either, that is why I am removing it.

PR #128, opened 2026-04-13
[MISC] 8.4 - Rootless K8s container

sinclert-canonical:
This PR updates the MySQL Router K8s charm to make the workload container a root-less one.

### DIFF
--- a/common/common/workload.py
+++ b/common/common/workload.py
@@ -47,7 +47,6 @@ class Workload:
         self._container = container_
         self._logrotate = logrotate_
         self._cos = cos_
-        self._router_data_directory = self._container.path("/var/lib/mysqlrouter")
         self._tls_key_file = self._container.router_config_directory / "custom-key.pem"
         self._tls_certificate_file = (
             self._container.router_config_directory / "custom-certificate.pem"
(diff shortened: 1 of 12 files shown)


common/common/workload.py
```diff
@@ -167,10 +166,8 @@ def _disable_router(self) -> None:
         logger.info("Disabling MySQL Router service")
         self._container.update_mysql_router_service(enabled=False)
         self._logrotate.disable()
-        self._container.router_config_directory.rmtree()
-        self._container.router_config_directory.mkdir()
-        self._router_data_directory.rmtree()
-        self._router_data_directory.mkdir()
+        self._container.router_config_directory.empty()
+        self._container.router_data_directory.empty()
```
sinclert-canonical, line 170: These are the changes triggering all the other Path refactors: without escalated privileges we cannot re-create the `var/lib/mysqlrouter` folder within the container. The best we can do is to empty it.

kubernetes/src/kubernetes_logrotate.py
```diff
@@ -15,11 +15,11 @@
...
     """logrotate implementation for k8s"""
 
-    _SYSTEM_USER = "mysql"
-
     def __init__(self, *, container_: common.container.Container):
         super().__init__(container_=container_)
-        self._logrotate_executor = self._container.path("/logrotate_executor.py")
+        self._logrotate_executor = self._container.path(
+            f"/home/{container_.system_user}/logrotate_executor.py"
+        )
```
sinclert-canonical, line 22: Exactly the same as with MySQL Server: files cannot be places in the root directory (/) anymore.

astrojuanlu (approved):
LGTM!

paulomach (approved):

PR #146, opened 2026-05-26
[MISC] Fix Allure report publication

sinclert-canonical:
This PR proposed a similar setup for the `gh-pages-beta` branch as the one in MySQL Server monorepo (see [branch](https://github.com/canonical/mysql-operators/tree/gh-pages-beta)).

All previously published allure reports have been deleted, as the split between 8.0 and 8.4 branches is not trivial (will require going thought GitHub Actions history and match with allure report creation timestamps). Furthermore, old reports do not provide much value anymore.


astrojuanlu (approved):
As discussed offline, if we ever want to bring back the history, we can check it out from old commits as needed 👍🏼

paulomach (approved):

PR #149, opened 2026-05-29
[DPE-10195] 8.0 - Introduce mysql-shell-client v1

sinclert-canonical:
This PR introduces the `mysql-shell-client` v1 library into the MySQL Router 8.0 operators.

### Approach

The `Shell` common class has been preserve, as a way to keep all the MySQL Shell interactions together. It would be the same as the  `MySQLBase` class within the MySQL operator charmlib. This class receives an executor instance, which is the one that gets built differently depending whether we are in the kubernetes or machines folders.


paulomach (approved):
Looking great

sinclert-canonical:
@astrojuanlu

common/common/mysql_shell/__init__.py
```diff
@@ -122,38 +54,24 @@ def _get_attributes(self, additional_attributes: dict = None) -> str:
...
-                name_pattern=name_pattern,
-                output_filepath=output_file.relative_to_container,
-            )
-        )
-        with output_file.open("r") as file:
-            rows = json.load(file)
-        output_file.unlink()
-        logger.debug(f"MySQL roles found for {name_pattern=}: {len(rows)}")
-        return {row[0] for row in rows}
+        roles = self._instance_client.search_instance_roles("%")
```
paulomach: @sinclert-canonical , is the hard coded `%` intentional? Should it be `name_pattern` instead?

sinclert-canonical: Good catch. Updated in commit https://github.com/canonical/mysql-router-operators/pull/149/commits/b8a4dd77658885d44d8ef27d90d78d75f40db2ad.

common/common/mysql_shell/__init__.py
```diff
@@ -34,86 +32,20 @@ class RouterUserInformation:
...
 class Shell:
     """MySQL Shell connected to MySQL cluster"""
 
-    _container: container.Container
-    _connection_info: "database_requires.CompleteConnectionInformation"
+    def __init__(self, executor: BaseExecutor) -> None:
+        """Initialize the shell connection"""
+        self._executor = executor
+        self._cluster_client = ClusterClient(self._executor)
+        self._instance_client = InstanceClient(self._executor, QueryQuoter())
```
astrojuanlu, line 42: nit: I'm usually a fan of this pattern, in case you want to consider it

```python
@dataclass
class Shell:
    _executor: BaseExecutor
    _cluster_client: ClusterClient
    _instance_client: InstanceClient

    @classmethod
    def from_executor(cls, executor: BaseExecutor) -> "Shell":
        cluster_client = ClusterClient(self._executor)
        instance_client = InstanceClient(self._executor, QueryQuoter())
        return cls(executor, cluster_client, instance_client)
```

astrojuanlu (approved):
Excellent 👏🏼 so much cleaner!

common/common/mysql_shell/__init__.py
sinclert-canonical: I do not think the `Shell` class should be a dataclass, as it holds complex logic. Furthermore, the only reason to define the `@classmethod` constructor is that dataclasses overload the initializer, so some other way to initialize the dataclass with custom arguments needs to be defined.

To me using `@dataclass` is a declaration of intent, and here there is no such intent.

astrojuanlu: Fair enough! I think `@dataclass` is an unfortunate misnomer, big fan of Glyph's "every class can be a dataclass" philosophy https://blog.glyph.im/2025/04/stop-writing-init-methods.html

PR #150, opened 2026-05-29
[DPE-10195] 8.4 - Introduce mysql-shell-client v1

sinclert-canonical:
This PR introduces the `mysql-shell-client` v1 library into the MySQL Router 8.4 operators.

### Approach

The `Shell` common class has been preserve, as a way to keep all the MySQL Shell interactions together. It would be the same as the  `MySQLBase` class within the MySQL operator charmlib. This class receives an executor instance, which is the one that gets built differently depending whether we are in the kubernetes or machines folders.


paulomach (approved):
Looking great

sinclert-canonical:
@astrojuanlu

common/common/mysql_shell/__init__.py
```diff
@@ -122,38 +54,24 @@ def _get_attributes(self, additional_attributes: dict = None) -> str:
...
-                name_pattern=name_pattern,
-                output_filepath=output_file.relative_to_container,
-            )
-        )
-        with output_file.open("r") as file:
-            rows = json.load(file)
-        output_file.unlink()
-        logger.debug(f"MySQL roles found for {name_pattern=}: {len(rows)}")
-        return {row[0] for row in rows}
+        roles = self._instance_client.search_instance_roles("%")
```
paulomach: https://github.com/canonical/mysql-router-operators/pull/149/changes#r3358686287

sinclert-canonical: Good catch. Updated in commit https://github.com/canonical/mysql-router-operators/pull/150/commits/fe68c2476f27e4fe444fc11bec9752cdf3ddec92.

astrojuanlu (approved):

PR #160, opened 2026-06-09
[MISC] 8.4 - Fix charmcraft build process

sinclert-canonical:
This PR fixes both charms build process by manually setting the `mysql-shell-client` version in their pyproject.toml. I left the original approach of _removing charmcraftlocal from the build process_ show-casing that it is not enough.

### Notice:
A better solution needs to be found, so that we do not update the dependencies from the `poetry.lock` at construction time. This is not only a security hole (the versions we see in the `poetry.lock` are not what get included within the charms), but a source of confusion for us.

### DIFF
--- a/kubernetes/charmcraft.yaml
+++ b/kubernetes/charmcraft.yaml
@@ -25,8 +25,6 @@ parts:
       python3 -m pip install --user --break-system-packages poetry-plugin-export==1.9.0
 
       ln -sf "$HOME/.local/bin/poetry" /usr/local/bin/poetry
-
-      python3 -m pip install --user --break-system-packages charmcraftlocal
   # "charm-poetry" part name is arbitrary; use for consistency
   # Avoid using "charm" part name since that has special meaning to charmcraft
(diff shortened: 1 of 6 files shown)


astrojuanlu (approved):
Given the circumstances, this seems to be the quickest way forward 👍🏼

astrojuanlu:
charmcraftlocal 0.4.0 no longer refreshes the `poetry.lock` (thanks @carlcsaposs-canonical !) so this is no longer needed.

See: https://github.com/canonical/mysql-router-operators/actions/runs/27201426723/job/80804633931?pr=158#step:11:470

carlcsaposs-canonical:
> no longer refreshes the `poetry.lock`

FYI it still modifies poetry.lock, but should now only affect the local packages (and not the dependencies of local packages). More details about why this is needed: https://github.com/canonical/charmcraftlocal#why-does-charmcraftlocal-need-to-be-called-in-charmcraftyaml

Issue #165, opened 2026-06-12
Mysql router dashboard use wrong metric name


chanchiwai-ray:
## Steps to reproduce

Deploy mysql-router and relate to COS and check the dashboard

## Expected behavior

The panel showing "no data" use wrong metric names: `route_connections_byte_to_server`; it misses the `nameSpace` = `mysqlrouter`. See https://github.com/rluisr/mysqlrouter_exporter/blob/master/gauges.go#L77 for example.

## Actual behavior

<img width="3821" height="1630" alt="Image" src="https://github.com/user-attachments/assets/470e103f-77b9-4915-be5c-66b8b4a584e4" />

## Versions

mysql router 8.0/edge (I suppose 8.4/edge is also affected)

## Log output

N/A

chanchiwai-ray:
closed by #166

PR #166, opened 2026-06-12
8.0 - Fix wrong metrics name in dashboard

chanchiwai-ray:
All metric names are prefixed with a namespace: "mysqlrouter".

## Issue

The metric name should be prefixed by `mysqlrouter`. See https://github.com/rluisr/mysqlrouter_exporter/blob/master/gauges.go#L77 for example

## Solution

Add the `mysqlrouter` prefix

Closes: #165


chanchiwai-ray:
After adding prefix

image

astrojuanlu (approved):
Thank you @chanchiwai-ray !

paulomach (approved):
Thank you @chanchiwai-ray

PR #167, opened 2026-06-12
8.4 - Fix wrong metrics name in dashboard

chanchiwai-ray:
All metric names are prefixed with a namespace: "mysqlrouter".

## Issue

The metric name should be prefixed by `mysqlrouter`. See https://github.com/rluisr/mysqlrouter_exporter/blob/master/gauges.go#L77 for example

## Solution

Add the `mysqlrouter` prefix


astrojuanlu (approved):

paulomach (approved):
Thank you @chanchiwai-ray

PR #138, opened 2026-05-20
[MISC] 8.4 - Update canonical/data-platform-workflows to v49

sinclert-canonical:
This PR bumps the Data Platform workflows to version 49, making use of the new security practices (`zizmor.yaml`).

Integration tests were adapted to MySQL Test app changes introduced in this [PR](https://github.com/canonical/mysql-test-app/pull/59).

---

Alternative to PR https://github.com/canonical/mysql-router-operators/pull/134.

PR #174, opened 2026-06-22
[DPE-10381] (8.4/edge) Fix Allure reports

### LINKED ISSUE #163: Fix Allure reports
They have been failing, see for instance https://github.com/canonical/mysql-router-operators/actions/runs/27318094146/job/80711880820

astrojuanlu:
## Issue

Sorry, this 1 more thing 🥲 

Fix gh-163.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -101,7 +101,7 @@ jobs:
         with:
           ref: gh-pages-beta
           path: repo/
-          persist-credentials: false
+          persist-credentials: true
       - name: Parse branch name
         id: branch
         shell: python


sinclert-canonical (commented):
I am surprised by this change.

It is true that the nightly workflow in the MySQL operators repository (see [code](https://github.com/canonical/mysql-operators/blob/e603b87c841f90e9dddb3b038ad8d311df7565fc/.github/workflows/nightly.yaml#L93-L97)) does not specify the `persist-recentials` flag, which takes a `true` value by default, but I  assumed this is because we missed it, and not because it is necessary. Is this required to push to the `gh-pages-beta` branch afterwards?

astrojuanlu:
> It is true that the nightly workflow in the MySQL operators repository (see [code](https://github.com/canonical/mysql-operators/blob/e603b87c841f90e9dddb3b038ad8d311df7565fc/.github/workflows/nightly.yaml#L93-L97)) does not specify the persist-recentials flag, which takes a true value by default, but I assumed this is because we missed it, and not because it is necessary.

I think it has to do with mysql-operators not having the `lint-workflows` job 🤔 which you did include in https://github.com/canonical/mysql-router-operators/pull/138/, and forced you to add `persist-credentials: false`.

> Is this required to push to the gh-pages-beta branch afterwards?

Yes it is, otherwise it fails:

https://github.com/canonical/mysql-router-operators/actions/runs/27924401491/job/82628954443

That's how PostgreSQL does it:

https://github.com/canonical/postgresql-k8s-operator/blob/211a992faecc70cfff03b23a94caf014cb35033e/.github/workflows/integration_test.yaml#L240-L245

sinclert-canonical (approved):

paulomach (approved):

PR #148, opened 2026-05-28
[DPE-9970] 8.4 - Build for Ubuntu 26.04 base

sinclert-canonical:
This PR adds the Ubuntu 26.04 base to the charm.

---

Depends on:
- [Juju] Releasing a new 26.04 charm-base image - [PR](https://github.com/juju/charm-base-images/pull/19).
- [O11y] Bumping loki-push v1 architecture fetching - [PR](https://github.com/canonical/loki-k8s-operator/pull/637).
- [MySQL] Releasing a new 26.04 charmed-mysql snap - [PR](https://github.com/canonical/charmed-mysql-snap/pull/108).
- [MySQL] Releasing a new 26.04 charmed-mysql rock - [PR](https://github.com/canonical/charmed-mysql-rock/pull/86).
- [MySQL] Releasing a new 26.04 mysql-test-app charm - [PR](https://github.com/canonical/mysql-test-app/pull/61).
- [MySQL] Migrating integration tests to Juju CLI - [PR](https://github.com/canonical/mysql-router-operators/pull/168).

PR #180, opened 2026-06-25
[DPE-10381] (8.4/edge) Fix Allure filter

astrojuanlu:
## Issue

Runners changed to 26.04 in #148.

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -63,7 +63,7 @@ jobs:
           )
           jobs = []
           for job in spread_jobs:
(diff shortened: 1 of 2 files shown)


sinclert-canonical (commented):
Thanks! I can't wait to get rid of this conditional from the workflows.

We need to change the `release_test.yaml` [workflow](https://github.com/canonical/mysql-router-operators/blob/793a68f18dc8b15e021bec5d71be3e5e7a6ac0b3/.github/workflows/release_test.yaml) as well, even if we are not using it just yet.

paulomach (approved):

sinclert-canonical (approved):

PR #94, opened 2026-02-21
[MISC] Migration to jubilant

paulomach:
What says in the tin!
Works on my machine™️ 
Early stages, don't review

PR #181, opened 2026-06-26
[MISC] 8.0 - Migrate K8s integration tests to Jubilant

sinclert-canonical:
This PR migrates the Kubernetes integration tests to Jubilant. Given that MySQL Router 8.0 still supports Juju 2 and Juju 3, the `jubilant-backports` package has been added to the project.

### Strategy
All integration tests have been **manually ported to Jubilant (no AI involved)**, naming each new integration test module similarly to its old integration test counterpart, adding the suffix `_new`. This has been done to ease PR review and avoid a confusing diff.

The helpers introduced in the `helpers_new.py` module are **exactly the same** as the ones present in MySQL operators repository, which have proved battle-tested in that repo test suite. Of course, MySQL Router has some unique ones:
- [check_router_metrics_endpoint](https://github.com/canonical/mysql-router-operators/blob/b2eb0e35f7990557d60d834ce8cdd3a972bec6b0/kubernetes/tests/integration/helpers_new.py#L27-L47) used in the exporter and exporter-with-tls tests.
- [get_unit_certificate_issuer](https://github.com/canonical/mysql-router-operators/blob/b2eb0e35f7990557d60d834ce8cdd3a972bec6b0/kubernetes/tests/integration/helpers_new.py#L130-L150) used in the tls and exporter-with-tls tests.

### Future work
- Do the same for VM integration tests.

---

This is a new take to PR https://github.com/canonical/mysql-router-operators/pull/94.


paulomach (approved):
same as 8.4 - thank you

kubernetes/tests/integration/helpers_new.py
```diff
@@ -0,0 +1,292 @@
...
+    unit_address = get_unit_address(juju, app_name, unit_name)
+
+    try:
+        for attempt in Retrying(
+            stop=stop_after_delay(2 * MINUTE_SECS),
+            wait=wait_fixed(10),
+            reraise=True,
+        ):
+            with attempt:
+                response = requests.get(f"http://{unit_address}:9152/metrics", stream=False)
```
astrojuanlu, line 38: Will this work on HA Canonical K8s @paulomach ?

kubernetes/tests/integration/integration/test_exporter.py
```diff
@@ -0,0 +1,114 @@
...
+    )
+    juju.deploy(
+        charm=GRAFANA_AGENT_APP_NAME,
+        app=GRAFANA_AGENT_APP_NAME,
+        base=ubuntu_base,
+        channel="1/stable",
+        num_units=1,
+    )
+
+    logging.info("Relating the applications")
```
astrojuanlu, line 63: I notice we're no longer waiting for the router to become Blocked, does this have any effect? (Same for TLS)

kubernetes/tests/integration/integration/test_expose_external.py
```diff
@@ -0,0 +1,208 @@
...
+
+            assert is_connection_possible(connection_config, **{"ssl_disabled": False})
+
+
+def check_connectivity(juju: Juju) -> None:
+    """Helper function to check connectivity"""
+    juju.config(
+        app=MYSQL_ROUTER_APP_NAME,
+        values={"expose-external": "false"},
+    )
```
astrojuanlu, line 177: Shouldn't we leave config changes outside of this method? Like the old test did

kubernetes/tests/integration/integration/test_log_rotation.py
```diff
@@ -0,0 +1,261 @@
...
+    router_app_leader = get_app_leader(juju, MYSQL_ROUTER_APP_NAME)
+
+    logging.info("Stopping the logrotate executor pebble service")
+    juju.ssh(
+        target=router_app_leader,
+        container=CONTAINER_NAME,
+        command="pebble stop logrotate_executor",
+    )
+
+    logging.info("Removing existing archive directory")
```
astrojuanlu, line 87: Aren't we missing an equivalent of the old `stop_running_flush_mysqlrouter_job`?

kubernetes/tests/integration/integration/test_log_rotation.py
```diff
@@ -0,0 +1,261 @@
...
+    """
+    pod_name = get_mysql_instance_label(unit_name)
+
+    with tempfile.NamedTemporaryFile(mode="w", dir=Path.home()) as temp_file:
+        temp_file.write(file_data)
+        temp_file.flush()
+
+        subprocess.check_call(
+            [
+                "microk8s",
```
astrojuanlu, line 253: dpe running in MicroK8s 👍🏼

kubernetes/tests/integration/integration/test_upgrade.py
```diff
@@ -0,0 +1,227 @@
...
+
+    router_status = juju.status().apps[MYSQL_ROUTER_APP_NAME]
+    router_unit_status = router_status.units[router_app_units[0]].workload_status
+
+    # Refresh will be incompatible on PR CI (not edge CI)
+    # since unreleased charm versions are always marked as incompatible
+    if router_unit_status.current == "blocked" and "incompatible" in router_unit_status.message:
+        logging.info("Application upgrade is blocked due to incompatibility")
+        juju.run(
+            unit=router_app_units[0],
```
astrojuanlu, line 113: I notice we aren't checking the refresh order anymore.

And also, I discovered a few days ago that, if unit 1 is the leader instead of unit 0, the resume-refresh task will be marked as failed, despite the refresh being successful ([internal chat](https://chat.canonical.com/canonical/pl/6h4bpxfarjrafg3kiwwwnfa51a)). It might be a good moment to account for this, either by detecting "bad" leader and failing with a helpful message, or actively catching the exception.

kubernetes/tests/integration/architecture.py
```diff
@@ -1,7 +1,9 @@
 # Copyright 2024 Canonical Ltd.
 # See LICENSE file for licensing details.
+
 import subprocess
 
-architecture = subprocess.run(
```
astrojuanlu, line 5: I think the currently blessed high level interface is `subprocess.run` https://docs.python.org/3/library/subprocess.html#subprocess.run:~:text=The%20recommended%20approach%20to%20invoking%20subprocesses%20is%20to%20use%20the%20run()%20function%20for%20all%20use%20cases%20it%20can%20handle

kubernetes/tox.ini
```diff
@@ -60,4 +60,4 @@ pass_env =
 commands_pre =
     poetry install --only integration
 commands =
-    poetry run pytest -v --tb native --log-cli-level=INFO -s --ignore={[vars]tests_path}/unit/ {posargs}
+    poetry run pytest -v -x --tb native --log-cli-level=INFO -s --ignore={[vars]tests_path}/unit/ {posargs}
```
astrojuanlu, line 63: Should we revert this on CI?

astrojuanlu (commented):
Thank you @sinclert-canonical 👏🏼 left a few comments

kubernetes/tests/integration/integration/test_exporter.py
sinclert-canonical: I greatly simplified those waits across the test suite. To me, it does not make sense to wait for the MySQL Router application to get into a `blocked` state, only to relate it to MySQL Server to have it resolved + moved to `active` state. We just relate it from the very beginning.

If we want to test whether the MySQL Router application reaches a `blocked` state unless a specific relation is created, we can do so in a specific test, not across the test suite.

kubernetes/tests/integration/integration/test_expose_external.py
sinclert-canonical: Good point.

This is a necessary change that I made, in order to make this testing [function](https://github.com/canonical/mysql-router-operators/blob/eab88e5ad92309dec28aa4201738e38942df9477/kubernetes/tests/integration/integration/test_expose_external_new.py#L172-L208) idempotent.

After extracting all the IP reachability assertions to a re-usable function, such function needs to run all the necessary actions / config changes so that the assertions succeed, independently of the previous actions / config changes. Otherwise, the function will depend on **external state**, and will likely fail when used in the wrong places.

kubernetes/tests/integration/integration/test_log_rotation.py
sinclert-canonical: Yes, I am removing those lines both in the K8s and VM Jubilant integration tests.

Those lines where added in order to stop any _running_ log rotation process, if there was any running at the time of the test execution. Not only is the chance extremely small (they run hourly), but the way Jubilant executes the stopping commands over SSH validates that the exit code is `0`, which will almost never be true, as the stopping command is not idempotent.

In short: it is not worth the trouble.

kubernetes/tests/integration/integration/test_upgrade.py
sinclert-canonical: As far as I understand, the old tests where never _checking_ the units order, but just ordering them by the number that they have after `/` (which is the same thing as the native `list.sort` method does).

---

Regarding your discovery:

That limitation is definitely critical to the upgrade / refresh process, and may explain the flaky behaviour of the upgrade tests. However, I am confused about which version of the refresh logic is affected. Your MM message show-cases a MySQL Server 8.4 charm, which runs refresh v3. This is MySQL Router 8.0, which runs refresh v2 (very similar to v3).

Did you encounter the same limitation here?

kubernetes/tests/integration/architecture.py
sinclert-canonical: All good to me. Will revert to `subprocess.run`.

kubernetes/tox.ini
sinclert-canonical: Quite the opposite.

This it the pytest `-x` / `--exitfirst` flag, which would be equivalent as marking every test within a specific Python module with the _abort-on-fail_ decorator. This is critical on modules that contain several tests, as, given the eventual consistency nature of Juju (i.e. lots of waits), we may be wasting a lot of time waiting for specific conditions that will never be fulfilled, given a crash on a previous test.

I clearly remember adding this to MySQL Server and how good it felt.

kubernetes/tests/integration/integration/test_upgrade.py
astrojuanlu: About the ordering, AFAIU the old test was running the `force-refresh-start` action on `refresh_order[0]`, which is the /2 unit. This test is running it on `router_app_units[0]`, which is /0. Is my understanding correct?

---

Oh, you're absolutely right, what I found is about Refresh v3 👍🏼

kubernetes/tests/integration/integration/test_expose_external.py
astrojuanlu: Fair enough, the nature of the helper has changed but that shouldn't be a blocker 👍🏼

kubernetes/tests/integration/integration/test_upgrade.py
sinclert-canonical: > Is my understanding correct?

Both the old code and the new code sort the units with the `reverse=True` flag, so, in both cases, we are targeting `mysql-router-k8s/2`.

kubernetes/tests/integration/architecture.py
sinclert-canonical: Done in commit https://github.com/canonical/mysql-router-operators/pull/181/commits/795993345fae6717fda8e7fc44babe78db94bd13.

kubernetes/tests/integration/integration/test_upgrade.py
astrojuanlu: I missed the `router_app_units.sort(reverse=True)` line 👍🏼 All good!

sinclert-canonical:
🗞️  Reverting the test modules `_new` suffix to ease PR review + merging.

kubernetes/tests/integration/helpers_new.py
paulomach: Should be the same as reaching the db endpoint, so for local ck8s yes.

Issue #24, opened 2025-07-07
Removing application msql-router-k8s resulted into an error state


hemanthnakkina:
## Steps to reproduce

1. Install juju+k8s cloud
2. juju deploy ./designate.txt --trust
3. Wait for all units in active and idle state
4. juju remove-application designate-mysql-router designate --no-prompt

[designate.txt](https://github.com/user-attachments/files/21094356/designate.txt)

I think rerunning the steps multiple times might reproduce the issue atleast once in the first 3 runs.

## Expected behavior

juju units designate and designate-mysql-router is removed

## Actual behavior

designate-mysql-router is in error state

```
designate-mysql-router/0*  error     idle   10.1.0.79          hook failed: "database-relation-broken"
designate/0*               blocked   idle   10.1.0.137         (dns-backend) integration missing
```

## Versions

Operating system: ubuntu 24.04

Juju CLI: 3.6.7

Juju agent: 3.6.7

mysql-k8s charm revision: 255
mysql-router-k8s charm revision: 748 

k8s: 1.32.6

## Log output

Juju debug log for designate-mysql-router: 
```
025-07-06T13:20:38.858Z [container-agent] 2025-07-06 13:20:38 INFO juju.worker.uniter.operation runhook.go:186 ran "backend-database-relation-joined" hook (via hook dispatching script: dispatch)
2025-07-06T13:20:40.495Z [container-agent] 2025-07-06 13:20:40 INFO juju.worker.uniter.operation runhook.go:186 ran "database-relation-joined" hook (via hook dispatching script: dispatch)
2025-07-06T13:20:42.048Z [container-agent] 2025-07-06 13:20:42 INFO juju.worker.uniter.operation runhook.go:186 ran "backend-database-relation-changed" hook (via hook dispatching script: dispatch)
2025-07-06T13:20:43.779Z [container-agent] 2025-07-06 13:20:43 INFO juju.worker.uniter.operation runhook.go:186 ran "database-relation-changed" hook (via hook dispatching script: dispatch)
2025-07-06T13:20:49.094Z [container-agent] 2025-07-06 13:20:49 INFO juju.worker.uniter.operation runhook.go:186 ran "upgrade-version-a-relation-changed" hook (via hook dispatching script: dispatch)
[...]

PR #183, opened 2026-06-30
[DPE-7684] (dpe) Avoid crashing on teardown if credentials are revoked

paulomach:
## Issue

See issue #24

## Solution

On teardown, if credentials are revoked, avoid breaking the charm

### DIFF
--- a/common/common/workload.py
+++ b/common/common/workload.py
@@ -15,6 +15,7 @@ import charm_refresh
 import ops
 import requests
 import tenacity
+from mysql_shell.executors.errors import ExecutionError
 
 from . import container, mysql_shell, server_exceptions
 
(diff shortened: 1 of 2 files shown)


common/common/workload.py
```diff
@@ -400,9 +401,20 @@ def reconcile(
...
+            router_not_in_cluster_set = (
+                self._router_id not in self.shell.get_routers_in_cluster_set()
+            )
+        except ExecutionError:
+            # Credentials may have been revoked by MySQL during teardown
+            logger.warning(
+                "Failed to query cluster set routers (credentials may have been revoked). "
+                "Disabling router"
+            )
+            router_not_in_cluster_set = True
```
sinclert-canonical: Shouldn't we put in the `try-except` block the smallest snippet that can crash?

```python3
try:
    cluster_set_routers = self.shell.get_routers_in_cluster_set()
except ExecutionError:
    cluster_set_routers = []
    logger.warning(
        "Failed to query cluster set routers (credentials may have been revoked). "
        "Disabling router"
    )

if any((
    is_charm_exposed == socket_file_exists,
    self._router_id not in cluster_set_routers,
)):
    ...
```

sinclert-canonical (commented):
Great finding!

Is there any _easy_ way to test against it? Any kind of charm that gets related / unrelated in a forceful manner perhaps?

paulomach:
> Great finding!
> 
> Is there any _easy_ way to test against it? Any kind of charm that gets related / unrelated in a forceful manner perhaps?

`terraform destroy` is the use case where that's is seem, usual for sunbeam

sinclert-canonical (approved):
As always, we need to port this to the `8.4/edge` branch as well.

common/common/workload.py
paulomach: 7e864369

paulomach:
> As always, we need to port this to the `8.4/edge` branch as well.

#186

astrojuanlu (approved):

PR #184, opened 2026-07-01
[MISC] 8.0 - Migrate VM integration tests to Jubilant

sinclert-canonical:
This PR migrates the Machines integration tests to Jubilant. Given that MySQL Router 8.0 still supports Juju 2 and Juju 3, the `jubilant-backports` package has been added to the project.

### Strategy
All integration tests have been **manually ported to Jubilant (no AI involved)**, naming each new integration test module similarly to its old integration test counterpart, adding the suffix `_new`. This has been done to ease PR review and avoid a confusing diff.

The helpers introduced in the `helpers_new.py` module are **exactly the same** as the ones present in MySQL operators repository, which have proved battle-tested in that repo test suite.

---

This is a new take to PR https://github.com/canonical/mysql-router-operators/pull/94.


sinclert-canonical:
There is something wrong with IPs in the HACluster tests. Investigating...

kubernetes/charmcraft.yaml
```diff
@@ -5,7 +5,6 @@ type: charm
 platforms:
   ubuntu@22.04:amd64:
   ubuntu@22.04:arm64:
-  ubuntu@22.04:s390x:
```
astrojuanlu, line 8: I assume this is temporary?

machines/tests/integration/integration/test_data_integrator.py
```diff
@@ -0,0 +1,198 @@
...
+    data_integrator_creds = juju.run(
+        unit=data_integrator_leader,
+        action="get-credentials",
+    )
+
+    databases = execute_queries_against_unit(
+        username=data_integrator_creds.results["mysql"]["username"],
+        password=data_integrator_creds.results["mysql"]["password"],
+        host=data_integrator_creds.results["mysql"]["endpoints"].split(",")[0].split(":")[0],
+        port=data_integrator_creds.results["mysql"]["endpoints"].split(",")[0].split(":")[1],
```
astrojuanlu, line 104:
```suggestion

    host, port = data_integrator_creds.results["mysql"]["endpoints"].split(",")[0].split(":")
    databases = execute_queries_against_unit(
        username=data_integrator_creds.results["mysql"]["username"],
        password=data_integrator_creds.results["mysql"]["password"],
        host=host,
        port=port,
```

?

machines/tests/integration/integration/test_data_integrator_new.py
```diff
@@ -0,0 +1,198 @@
...
+        stop=stop_after_delay(5 * MINUTE_SECS),
+        wait=wait_fixed(10),
+        reraise=True,
+    ):
+        with attempt:
+            assert "CN = Test CA" in (
+                get_unit_certificate_issuer(juju, router_leader, address=mysql_addr)
+            )
+
+    _ = execute_queries_against_unit(
```
astrojuanlu:
```suggestion
    execute_queries_against_unit(
```

machines/tests/integration/integration/test_data_integrator.py
```diff
@@ -0,0 +1,198 @@
...
+    data_integrator_creds = juju.run(
+        unit=data_integrator_leader,
+        action="get-credentials",
+    )
+
+    mysql_user = data_integrator_creds.results["mysql"]["username"]
+    mysql_pass = data_integrator_creds.results["mysql"]["password"]
+    mysql_addr = data_integrator_creds.results["mysql"]["endpoints"].split(",")[0]
+
+    logging.info("Ensuring no data exists in the test database")
```
astrojuanlu, line 125: Similar `host, port = mysql_addr.split(":")` that can then be reused further down?

machines/tests/integration/integration/test_exporter.py
```diff
@@ -0,0 +1,102 @@
...
+        base=ubuntu_base,
+        channel="8.0/edge",
+        config={"profile": "testing"},
+        num_units=1,
+    )
+    juju.deploy(
+        charm=charm,
+        app=MYSQL_ROUTER_APP_NAME,
+        base=ubuntu_base,
+        num_units=1,
```
astrojuanlu, line 38: Interestingly, the old test deployed 0 units of router and grafana apps, any idea why?

machines/tests/integration/integration/test_hacluster.py
```diff
@@ -0,0 +1,371 @@
...
+        delay=5.0,
+    )
+
+    data_integrator_leader = get_app_leader(juju, DATA_INTEGRATOR_APP_NAME)
+    data_integrator_creds = juju.run(unit=data_integrator_leader, action="get-credentials")
+
+    mysql_user = data_integrator_creds.results["mysql"]["username"]
+    mysql_pass = data_integrator_creds.results["mysql"]["password"]
+    mysql_host = data_integrator_creds.results["mysql"]["endpoints"].split(",")[0].split(":")[0]
+    mysql_port = data_integrator_creds.results["mysql"]["endpoints"].split(",")[0].split(":")[1]
```
astrojuanlu, line 97:
```suggestion
    mysql_host, mysql_port = data_integrator_creds.results["mysql"]["endpoints"].split(",")[0].split(":")
```

?

machines/tests/integration/integration/test_hacluster_new.py
```diff
@@ -0,0 +1,371 @@
...
+    check_server_accessible_virtual_ip(juju, MYSQL_ROUTER_VIP)
+
+
+def test_ha_cluster_failover(juju: Juju, ubuntu_base: str) -> None:
+    """Test the failover of the ha-cluster leader."""
+    ha_cluster_leader = get_app_leader(juju, HA_CLUSTER_APP_NAME)
+    ha_cluster_address = get_unit_machine_address(juju, HA_CLUSTER_APP_NAME, ha_cluster_leader)
+
+    logging.info("Stopping HACluster LXC container")
+    subprocess.check_call(["lxc", "stop", ha_cluster_address])
```
astrojuanlu:
```suggestion
    subprocess.run(["lxc", "stop", ha_cluster_address])
```

https://docs.python.org/3/library/subprocess.html#:~:text=The%20recommended%20approach%20to%20invoking%20subprocesses%20is%20to%20use%20the%20run()%20function%20for%20all%20use%20cases%20it%20can%20handle (and anywhere else using something other than `subprocess.run`)

machines/tests/integration/integration/test_hacluster.py
```diff
@@ -0,0 +1,371 @@
...
+
+    logging.info("Ensuring that VIP is not the data-integrator endpoint hostname")
+    data_integrator_leader = get_app_leader(juju, DATA_INTEGRATOR_APP_NAME)
+    data_integrator_creds = juju.run(unit=data_integrator_leader, action="get-credentials")
+
+    mysql_host = data_integrator_creds.results["mysql"]["endpoints"].split(",")[0].split(":")[0]
+    assert mysql_host != MYSQL_ROUTER_VIP
+
+
+def check_server_accessible_virtual_ip(juju: Juju, vip: str) -> None:
```
astrojuanlu, line 320: Type checkers complaining that some calls to this function pass the global MYSQL_ROUTER_VIP which might be none. In normal operation this is not an issue, but sometimes one wants to do `tox -e integration -- tests/.../test_hacluster.py -k only_this_test` and it's annoying to have a weird error because some earlier test didn't run (affects all our tests using global variables)

Maybe wrap the calls with `if not MYSQL_ROUTER_VIP: raise`?

machines/tests/integration/integration/test_log_rotation.py
```diff
@@ -0,0 +1,230 @@
...
+    router_leader = get_app_leader(juju, MYSQL_ROUTER_APP_NAME)
+
+    logging.info("Removing the cron file")
+    delete_unit_file(
+        juju=juju,
+        unit_name=router_leader,
+        file_path="/etc/cron.d/flush_mysqlrouter_logs",
+    )
+
+    logging.info("Removing existing archive directory")
```
astrojuanlu, line 77: Do we need an equivalent to `stop_running_flush_mysqlrouter_cronjobs` right before this step?

machines/tests/integration/integration/test_upgrade.py
```diff
@@ -0,0 +1,211 @@
...
+
+    model_status = juju.status()
+    router_status = model_status.apps[MYSQL_ROUTER_APP_NAME].app_status
+
+    # Refresh will be incompatible on PR CI (not edge CI)
+    # since unreleased charm versions are always marked as incompatible
+    if router_status.current == "blocked" and "incompatible" in router_status.message:
+        logging.info("Application upgrade is blocked due to incompatibility")
+        juju.run(
+            unit=router_app_units[0],
```
astrojuanlu, line 107: Seems we're no longer following refresh order, like in the old test. Any reason why?

machines/tests/integration/architecture.py
```diff
@@ -1,7 +1,9 @@
 # Copyright 2024 Canonical Ltd.
 # See LICENSE file for licensing details.
+
 import subprocess
 
-architecture = subprocess.run(
-    ["dpkg", "--print-architecture"], capture_output=True, check=True, encoding="utf-8"
-).stdout.strip()
+architecture = subprocess.check_output(
```
astrojuanlu: Should still be `subprocess.run` if possible

machines/charmcraft.yaml
```diff
@@ -5,7 +5,6 @@ type: charm
 platforms:
   ubuntu@22.04:amd64:
   ubuntu@22.04:arm64:
-  ubuntu@22.04:s390x:
```
astrojuanlu, line 8: Is this temporary?

astrojuanlu (commented):
Thank you @sinclert-canonical ! Left some comments

machines/tests/integration/integration/test_log_rotation.py
astrojuanlu: Clarified in #181

machines/tests/integration/integration/test_upgrade.py
astrojuanlu: Clarified in #181

kubernetes/charmcraft.yaml
sinclert-canonical: Yes, this is only here because I am iterating on the HA-Cluster integration test (see [commit](https://github.com/canonical/mysql-router-operators/pull/184/commits/2fec3b396a8a2033f1e4e85a37602367b6c8fb32)) and I would like to speed up the feedback loop.

machines/tests/integration/integration/test_exporter.py
sinclert-canonical: Yes. Both `mysql-router` and `grafana-agent` are _subordinate_ charms, so, technically, one cannot specify the number of units when deployed, as it depends on the _principal_ charm units.

From a design point of view, passing `0` or `None` to the number of units makes the most sense, and the OpstTest library we used to orchestrate the tests allow this. However, Jubilant is way stricter about the value passed to the `num_units` argument, raising a ValueError if the value is not a positive integer.

Therefore, it is either:
- A) We avoid such argument altogether (it is optional).
- B) We set its default value explicitly.

machines/charmcraft.yaml
sinclert-canonical: Yes, this is only here because I am iterating on the HA-Cluster integration test (see [commit](https://github.com/canonical/mysql-router-operators/pull/184/commits/2fec3b396a8a2033f1e4e85a37602367b6c8fb32)) and I would like to speed up the feedback loop.

machines/tests/integration/integration/test_exporter.py
astrojuanlu: Thanks for the explanation, that makes sense. I'd be in favour of (A) but it's a nitpick

machines/tests/integration/integration/test_data_integrator.py
sinclert-canonical: I dislike multi-variable assignment in Python.

machines/tests/integration/integration/test_data_integrator_new.py
sinclert-canonical: Addressed in commit https://github.com/canonical/mysql-router-operators/pull/184/commits/1753489e539f01e74f337a851d346abaacfea368.

machines/tests/integration/integration/test_data_integrator.py
sinclert-canonical: I dislike multi-variable assignment in Python, but at least I am extracting the MySQL host and port in a single place, instead of all over the test, in commit https://github.com/canonical/mysql-router-operators/pull/184/commits/1753489e539f01e74f337a851d346abaacfea368.

machines/tests/integration/integration/test_hacluster.py
sinclert-canonical: I dislike multi-variable assignment in Python.

machines/tests/integration/integration/test_hacluster.py
sinclert-canonical: Addressed in commit https://github.com/canonical/mysql-router-operators/pull/184/commits/1753489e539f01e74f337a851d346abaacfea368.

machines/tests/integration/architecture.py
sinclert-canonical: Addressed in commit https://github.com/canonical/mysql-router-operators/pull/184/commits/1753489e539f01e74f337a851d346abaacfea368.

machines/tests/integration/integration/test_hacluster_new.py
```diff
@@ -0,0 +1,387 @@
...
+def test_ha_cluster_failover(juju: Juju, ubuntu_base: str) -> None:
+    """Test the failover of the ha-cluster leader."""
+    data_integrator_leader = get_app_leader(juju, DATA_INTEGRATOR_APP_NAME)
+    data_integrator_units = get_app_units(juju, DATA_INTEGRATOR_APP_NAME)
+    data_integrator_units.remove(data_integrator_leader)
+
+    machine_id = get_unit_machine_id(juju, DATA_INTEGRATOR_APP_NAME, data_integrator_units[0])
+
+    logging.info("Stopping LXC container")
+    subprocess.check_call(["lxc", "stop", machine_id])
```
paulomach: I know this is in the original test, but we should be able to `lxc stop --force`  and not have to wait until `unknown` status

machines/tests/integration/integration/test_upgrade_new.py
```diff
@@ -0,0 +1,211 @@
...
+        )
+
+    logging.info("Wait for first unit to upgrade")
+    juju.wait(
+        ready=jubilant_backports.all_agents_idle,
+        timeout=5 * MINUTE_SECS,
+    )
+
+    # If leader is next to refresh, charm will be killed before action can succeed
+    with suppress(jubilant_backports.TaskError):
```
paulomach: Isn't this only valid for k8s charms (as the pod get rolled)?

machines/tests/integration/helpers_new.py
paulomach, line 138: nit: is there a need for having `cluster_set`? quick grep does not found any usage

paulomach (approved):
Looks excellent, though upgrade tests are consistently failing - fix those and we are good to go. Couple questions. Reach out if help for debugging is needed

machines/tests/integration/integration/test_hacluster_new.py
sinclert-canonical: Addressed in commit https://github.com/canonical/mysql-router-operators/pull/184/commits/86f4683054996b7790dbd1f4a5432baa16d3d63f.

machines/tests/integration/integration/test_upgrade_new.py
sinclert-canonical: You are right, too much copy+paste. Addressed in commit https://github.com/canonical/mysql-router-operators/pull/184/commits/86f4683054996b7790dbd1f4a5432baa16d3d63f.

machines/tests/integration/helpers_new.py
sinclert-canonical: You are right, it is not used anywhere. Still, I am inclined to leave it to match its K8s helper counterpart.

sinclert-canonical:
> Looks excellent, though upgrade tests are consistently failing.

Yes. I noticed that the machine upgrade tests have the logic of their _fail-and-rollback_ upgrade test differently.

In the machine upgrade test the rollback needs to be done to a hand-crafted _valid charm_, whose snap revisions / workload version has been setup by us. This is different from the K8s upgrade test though, where we can just rollback to the charm that gets packed during the build process (i.e. no hand-crafting involved).

sinclert-canonical:
🗞️  Reverting the test modules `_new` suffix to ease PR review + merging.

PR #191, opened 2026-07-16
[DPE-10198] 8.4 - Expose connection-sharing config

sinclert-canonical:
This PR exposes a `connection-sharing` config option to enable/disable MySQL Router 8.4+ connection-sharing [feature](https://dev.mysql.com/doc/mysql-router/8.4/en/mysql-router-general-features-connection-sharing.html). The proposed config options have been available since version 8.0.32, but the feature **does not** work in 8.0 versions.

### Decisions:

- Set [idle-timeout](https://dev.mysql.com/doc/mysql-router/8.4/en/mysql-router-conf-options.html#option_mysqlrouter_idle_timeout_connection-pool) to `60` seconds. Recommended value by Miguel Araujo (Oracle).
- Set [max-idle-server-connections](https://dev.mysql.com/doc/mysql-router/8.4/en/mysql-router-conf-options.html#option_mysqlrouter_max_idle_server_connections_connection-pool) to `1000`. Arbitrary high enough value proposed by Paulo Machado (MySQL).

### Additional changes

I needed to harmonize the `execute_queries_in_unit` test helper between K8s and VM, so that it is able to receive both a host and a port value, needed when directing the connections through MySQL Router.

### DIFF
--- a/common/common/workload.py
+++ b/common/common/workload.py
@@ -280,6 +280,10 @@ class RunningWorkload(Workload):
                 "routing:bootstrap_rw.connection_sharing=1",
                 "--conf-set-option",
                 "routing:bootstrap_ro.connection_sharing=1",
+                "--conf-set-option",
+                "connection_pool.idle_timeout=60",
+                "--conf-set-option",
+                "connection_pool.max_idle_server_connections=1000",
             ])
         return command
 
@@ -352,7 +356,7 @@ class RunningWorkload(Workload):
         if not self._container.router_config_file.exists():
             return False
         return self._parse_connection_sharing_from_config(
-            self._container.router_config_file.read_text()
+            self._container.router_config_file.read_text(),
         )
 
     @property
@@ -361,7 +365,9 @@ class RunningWorkload(Workload):
 
         During bootstrap, MySQL Router creates a config file which includes a generated username.
         """
-        return self._parse_username_from_config(self._container.router_config_file.read_text())
+        return self._parse_username_from_config(
+            self._container.router_config_file.read_text(),
+        )
 
     def _restart(self, *, event, tls: bool) -> None:
         """Restart MySQL Router to enable or disable TLS."""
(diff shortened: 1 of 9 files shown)


kubernetes/tests/integration/integration/test_conn_sharing.py
```diff
@@ -0,0 +1,148 @@
...
+
+        # This sleep time needs to be higher than the value of the
+        # MySQL Router `connection_sharing_delay` (defaults to `1`).
+        # Otherwise, the connections will not be reused
+        time.sleep(2)
+
+    for connection in connections:
+        connection.close()
+
+    return set(connection_ids)
```
paulomach, line 148: nice and tidy

common/common/workload.py
```diff
@@ -396,6 +431,18 @@ def reconcile(
...
+        # Check if connection sharing config has changed
+        connection_sharing_config = bool(self._charm.config.get("connection-sharing"))
+        connection_sharing_changed = all((
+            self._container.mysql_router_service_enabled,
+            self._router_connection_sharing_enabled != connection_sharing_config,
+        ))
+        if connection_sharing_changed:
+            logger.info(
+                f"Connection sharing config changed to {connection_sharing_config}. "
+                "Restarting router"
```
paulomach: just to confirm, we are not just restarting, but rebootstrapping it

paulomach (approved):
Minor question - nice solution for the test.

common/common/workload.py
sinclert-canonical: Aren't we disabling + enabling the Router a few lines below, depending on this value?

common/common/workload.py
```diff
@@ -273,6 +273,19 @@ def _get_bootstrap_command(
...
+        if self._charm.config.get("connection-sharing"):
+            command.extend([
+                "--conf-set-option",
+                "routing:bootstrap_rw.connection_sharing=1",
+                "--conf-set-option",
+                "routing:bootstrap_ro.connection_sharing=1",
+                "--conf-set-option",
+                "connection_pool.idle_timeout=60",
+                "--conf-set-option",
+                "connection_pool.max_idle_server_connections=1000",
```
astrojuanlu, line 286: Should we make these module constants at least?


```suggestion
                f"connection_pool.idle_timeout={CONNECTION_POOL_IDLE_TIMEOUT}",
                "--conf-set-option",
                f"connection_pool.max_idle_server_connections={CONNECTION_POOL_MAX_IDLE_SERVER_CONNECTIONS}",
```

so that we can also document them separately

astrojuanlu (approved):

common/common/workload.py
sinclert-canonical: You are absolutely right, the `disable` / `enable` tuple confused me.

common/common/workload.py
sinclert-canonical: Sadly, this repository does not follow the same structure as MySQL Server (i.e. no top-level `constants.py`), but that is something we can create in a follow-up PR. We can also extract some of the values from [here](https://github.com/canonical/mysql-router-operators/blob/01a0ea8aacf33f27d4196e363b4beb1a738b9ea6/common/common/workload.py#L255-L275).

For now, we can always refer to the official MySQL Router documentation.

PR #193, opened 2026-07-21
[DPE-10198] 8.0 - Fix connectivity-change check

sinclert-canonical:
This PR fixes a bug where, when an application passes the `external_node_connectivity` flag to the DP relation data, MySQL Router is unable to detect that it should restart itself after going through a Juju refresh. The existing condition is not complex enough to capture all the possible cases where we wanted to do a restart.

### Additional changes
We also needed to harmonize K8s and VM `get_unit_certificate_issuer` test helpers, now that MySQL Test app asks for the MySQL Router to be externally connectable (no socket available).

### DIFF
--- a/machines/tests/integration/integration/test_exporter_with_tls.py
+++ b/machines/tests/integration/integration/test_exporter_with_tls.py
@@ -13,6 +13,7 @@ from ..helpers_new import (
     MINUTE_SECS,
     check_router_metrics_endpoint,
     get_app_leader,
+    get_unit_address,
     get_unit_certificate_issuer,
     wait_for_apps_status,
 )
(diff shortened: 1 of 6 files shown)


machines/tests/integration/integration/test_exporter_with_tls.py
```diff
@@ -95,7 +94,8 @@ def test_exporter_endpoint(juju: Juju, charm: str, ubuntu_base: str) -> None:
     )
 
     router_leader = get_app_leader(juju, MYSQL_ROUTER_APP_NAME)
-    router_issuer = get_unit_certificate_issuer(juju, router_leader, socket=MYSQL_ROUTER_SOCKET)
+    router_address = get_unit_address(juju, MYSQL_ROUTER_APP_NAME, router_leader)
+    router_issuer = get_unit_certificate_issuer(juju, router_leader, router_address, 6446)
```
astrojuanlu, line 98:
```suggestion
    router_issuer = get_unit_certificate_issuer(juju, router_leader, router_address, CERT_ISSUER_PORT)
```

?

astrojuanlu (approved):
Made just 1 comment, it applies everywhere.

Otherwise LGTM!

Soundarya03 (approved):

machines/tests/integration/integration/test_exporter_with_tls.py
sinclert-canonical: Port `6446` is the standard port that MySQL Router uses to serve connections. Nothing specific to certificates

PR #194, opened 2026-07-22
[DPE-10198] 8.4 - Fix connectivity-change check

sinclert-canonical:
This PR fixes a bug where, when an application passes the `external_node_connectivity` flag to the DP relation data, MySQL Router is unable to detect that it should restart itself after going through a Juju refresh. The existing condition is not complex enough to capture all the possible cases where we wanted to do a restart.

### Additional changes
We also needed to harmonize K8s and VM `get_unit_certificate_issuer` test helpers, now that MySQL Test app asks for the MySQL Router to be externally connectable (no socket available).

### DIFF
--- a/machines/tests/integration/helpers_new.py
+++ b/machines/tests/integration/helpers_new.py
@@ -119,13 +119,11 @@ def get_unit_address(juju: Juju, app_name: str, unit_name: str) -> str:
     raise Exception("No application unit found")
 
 
-def get_unit_certificate_issuer(juju: Juju, unit_name: str, address=None, socket=None) -> str:
+def get_unit_certificate_issuer(juju: Juju, unit_name: str, host: str, port: int) -> str:
     """Get the TLS certificate issuer string."""
-    connect_args = f"-connect {address}" if address else f"-unix {socket}"
-
     output = juju.ssh(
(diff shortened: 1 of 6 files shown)


astrojuanlu (approved):
Leaving it approved, same comment as the companion PR

machines/tests/integration/helpers_new.py
```diff
@@ -119,13 +119,11 @@ def get_unit_address(juju: Juju, app_name: str, unit_name: str) -> str:
     raise Exception("No application unit found")
```
Soundarya03, line 121: Probably unrelated to this PR, but why do we have `helpers.py` and `helpers_new.py`?

Soundarya03 (approved):
LGTM!

machines/tests/integration/helpers_new.py
sinclert-canonical: The `helpers.py` module is the one we used to have when all integration tests used the `OpsTest` framework. Since the migration to Jubilant, all the Jubilant-styled helpers were added to a `helpers_new.py` module, to ease the diff shown during the PR review process.

We can unify them. No need to have both.

PR #199, opened 2026-08-14
WIP [DPE-10154] 8.4 - address tiobe issues in router

Soundarya03:
## Issue
This PR addresses issues highlighted in the TIOBE scan of the mysql-router operator (in both k8s and machine modules), in the 8.4 version.

### Summary of changes:
1. Adds coverage report generation to unit test run, tiobe scan broken without it
2.


.github/workflows/tiobe_scan_pr.yaml
```diff
@@ -0,0 +1,74 @@
+# Copyright 2026 Canonical Ltd.
+# See LICENSE file for licensing details.
+#
+# TEMPORARY workflow to test TICS scan changes on the fix/tiobe-machine-issues
+# branch before merging.
+# TODO: Remove this file before merge.
```
Soundarya03: TODO: this file should be removed before merging. It's been added temporarily to display results after applying chnages from this branch. @Soundarya03 to resolve before merging.

Soundarya03: Please also ignore the lint job failing on this workflow, as it's going to be removed anyways.

.github/workflows/tiobe_scan_pr.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+      - name: Coverage reports for TICS
+        run: |
+          mkdir -p ${{ matrix.operator }}/.cover
+          cp ${{ matrix.operator }}/coverage.xml ${{ matrix.operator }}/.cover/cobertura.xml
+
+      - name: Install TICS dependencies
+        run: |
+          . .venv/bin/activate
+          pip install --target=common "mysql-shell-client[contrib]>=1.0.0"
```
sinclert-canonical: How is this a dependency for TIOBE?

In other words: isn't it this dependency specified in the correct `pyproject.toml` files?

kubernetes/pyproject.toml
```diff
@@ -59,6 +59,8 @@ ops = "^2.9.0, <2.10"
 lightkube = "^0.17.2"
 tenacity = "^9.1.2"
 charm-refresh = "^3.1.0.2"
+# packaging.pylock (PEP 771) added in 25.1; required by tox at import time
+packaging = ">=25.1"
```
sinclert-canonical: I am confused by this change, as we have been able to share code between folders + build the charms without needing this package. Why do we needed it now?

machines/pyproject.toml
```diff
@@ -56,6 +56,8 @@ ops = "^2.9.0, <2.10"
 tenacity = "^9.1.2"
 charm-refresh = "^3.1.0.2"
 charmlibs-snap = "^1.0.1"
+# packaging.pylock (PEP 771) added in 25.1; required by tox at import time
+packaging = ">=25.1"
```
sinclert-canonical: I am confused by this change, as we have been able to share code between folders + build the charms without needing this package. Why do we needed it now?

sinclert-canonical (commented):
I have mixed feelings about the tests movement. In principle, there should be a way to provide the TIOBE score, leaving the unit tests where they are (similarly to how MySQL Server does it), even if there are 2 duplicated ones.

Is there an approach that would allow us to bump the TIOBE score, while avoiding the code shuffling?

kubernetes/pyproject.toml
Soundarya03: Hey Sinclert, thanks for flagging. This change was needed temporarily in order to get the temporary tiobe scan job on this branch working, but I can remove this now

Soundarya03: To be specific, I was using a self-hosted runner here and it was using a stale pckaging without pylock.
Nevertheless, this should now be reverted

Soundarya03:
> I have mixed feelings about the tests movement. In principle, there should be a way to provide the TIOBE score, leaving the unit tests where they are (similarly to how MySQL Server does it), even if there are 2 duplicated ones.
> 
> Is there an approach that would allow us to bump the TIOBE score, while avoiding the code shuffling?

Hmm, I see.
My rationale was:

1. The common module test coverage is duplicated across both substrates (byte-for-byte identical test_workload.py and test_architecture.py). This clearly doesn't test any substrate-specific behaviour, thus can be moved to the tests under the common folder. Also, if more pure-common tests are added in future, they have a natural home.
2. The scenario tests indeed vary between substrates ,and are different. But as they cover the common module, and the good practice is generally to have tests live next to the code they actually test is a best practice, I thought of moving these under common module too. Further, this makes the common module self-contained.

But I see your point. I can revert this re-org, and I an certainly add common module coverage without this shuffle. 

I'll let @paulomach confirm before reverting the re-org.

paulomach:
> > I have mixed feelings about the tests movement. In principle, there should be a way to provide the TIOBE score, leaving the unit tests where they are (similarly to how MySQL Server does it), even if there are 2 duplicated ones.
> > Is there an approach that would allow us to bump the TIOBE score, while avoiding the code shuffling?
> 
> Hmm, I see. My rationale was:
> 
>     1. The common module test coverage is duplicated across both substrates (byte-for-byte identical test_workload.py and test_architecture.py). This clearly doesn't test any substrate-specific behaviour, thus can be moved to the tests under the common folder. Also, if more pure-common tests are added in future, they have a natural home.
> 
>     2. The scenario tests indeed vary between substrates ,and are different. But as they cover the common module, and the good practice is generally to have tests live next to the code they actually test is a best practice, I thought of moving these under common module too. Further, this makes the common module self-contained.
> 
> 
> But I see your point. I can revert this re-org, and I an certainly add common module coverage without this shuffle.
> 
> I'll let @paulomach confirm before reverting the re-org.

I was a little confused about the move also, but I understand your reasoning. While not wrong I'd rather keep as it was.

Soundarya03:
> While not wrong I'd rather keep as it was.

Got it, thanks for the confirmation @paulomach 
I will make the changes and notify reviewers again when ready

.github/workflows/tiobe_scan_pr.yaml
Soundarya03: Removed, scanning is being triggered using a temporary PR #232

astrojuanlu (approved):
LGTM!

paulomach (approved):
Lgtm

kubernetes/tox.ini
```diff
@@ -51,7 +53,12 @@ description = Run unit tests
 commands_pre =
     poetry install --only main,charm-libs,unit
 commands =
-    poetry run pytest --numprocesses auto --cov=src --ignore={[vars]tests_path}/integration/ {posargs}
+    # Substrate unit tests -> coverage for both kubernetes/src and common/common.
+    # The substrate tests already import & exercise `common.*` at runtime, so a single run measures both modules.
+    poetry run pytest --numprocesses=auto --cov=src --cov={[vars]common_src_path} --cov-report=xml --ignore={[vars]tests_path}/integration/ {posargs}
```
sinclert-canonical: This command is getting out of hand, better to split it by lines. Please use the `{[vars]src_path}` shard variable when listing the coverage paths.

```diff
- poetry run pytest --numprocesses=auto --cov=src --cov={[vars]common_src_path} --cov-report=xml --ignore={[vars]tests_path}/integration/ {posargs}
+ poetry run pytest \
+    --numprocesses=auto \
+    --ignore={[vars]tests_path}/integration/ \
+    --cov={[vars]src_path} \
+    --cov={[vars]common_src_path} \
+    --cov-report=xml \
+    {posargs}
```

machines/tox.ini
```diff
@@ -50,7 +52,12 @@ description = Run unit tests
 commands_pre =
     poetry install --only main,charm-libs,unit
 commands =
-    poetry run pytest --numprocesses auto --cov=src --ignore={[vars]tests_path}/integration/ {posargs}
+    # Substrate unit tests -> coverage for both machines/src and common/common.
+    # The substrate tests already import & exercise `common.*` at runtime, so a single run measures both modules.
+    poetry run pytest --numprocesses=auto --cov=src --cov={[vars]common_src_path} --cov-report=xml --ignore={[vars]tests_path}/integration/ {posargs}
```
sinclert-canonical: This command is getting out of hand, better to split it by lines. Please use the `{[vars]src_path}` shard variable when listing the coverage paths.

```diff
- poetry run pytest --numprocesses=auto --cov=src --cov={[vars]common_src_path} --cov-report=xml --ignore={[vars]tests_path}/integration/ {posargs}
+ poetry run pytest \
+    --numprocesses=auto \
+    --ignore={[vars]tests_path}/integration/ \
+    --cov={[vars]src_path} \
+    --cov={[vars]common_src_path} \
+    --cov-report=xml \
+    {posargs}
```

sinclert-canonical (approved):
Thanks for addressing all the feedback! Left a couple of style improvements.

kubernetes/tox.ini
Soundarya03: Makes sense, thanks

PR #202, opened 2026-08-18
[DPE-10897] Fix and re-enable upgrade test on machines

astrojuanlu:
## Problem

The `test_upgrade.py` integration test fails after a charm upgrade because:

1. The mysqlrouter service crashes repeatedly after a snap refresh, hitting systemd's `StartLimitBurst` limit. `snap start` then fails with "Start request repeated too quickly", crashing the charm.
2. `wait_until_mysql_router_ready` raises `AssertionError` after 30 seconds if it can't connect to the router, crashing the charm during a `refresh-v-three` hook. This puts the unit in error state, preventing the router from being restarted.
3. After the crash and hook retry, `reconcile()` sees `mysql_router_service_enabled = True` (snap start succeeded) and skips re-bootstrapping, even though the router process is not actually listening on port 6446. The router stays in a broken state indefinitely.

## Changes

0. Build on top of @sinclert-canonical's #200, and revert #201.

1. **Retry `snap start`** (`machines/src/snap.py`) — wrap `snap start` in `tenacity.Retrying` (60-second timeout, 10-second intervals) to let systemd's rate-limit window expire and retry the start.

2. **Don't crash when the router is not ready** (`machines/src/charm.py`) — instead of raising `AssertionError` from `wait_until_mysql_router_ready` (which puts the unit in error state and stops hooks from firing), raise a `server_exceptions.Error` carrying a `WaitingStatus`. The existing `except server_exceptions.Error` handler in `abstract_charm.reconcile` catches it, sets the unit to `waiting`, and lets `update-status` keep firing. On the next hook, `_router_running_correctly` detects the port is down and re-bootstraps the router. This avoids the need for `juju resolved` — the charm recovers automatically.

3. **Verify router port in `_router_running_correctly`** (`common/common/workload.py`) — for machines TCP mode, check that the router is actually listening by calling `wait_until_mysql_router_ready` (wrapped in try/except `server_exceptions.Error` → return `False`). After a snap refresh, the service may be "enabled" without binding to the port. If the check fails, `reconcile()` will disable and re-enable the router, re-bootstrapping it with a fresh config.

### DIFF
--- a/common/common/workload.py
+++ b/common/common/workload.py
@@ -433,7 +433,7 @@ class RunningWorkload(Workload):
                 return False
             try:
                 self._charm.wait_until_mysql_router_ready(event=event)
-            except AssertionError:
+            except server_exceptions.Error:
                 return False
             return True
         else:
--- a/common/common/workload.py
+++ b/common/common/workload.py
@@ -427,7 +427,13 @@ class RunningWorkload(Workload):
         if charm_port_exposed is None:
             return self._container.mysql_router_service_enabled
         elif charm_port_exposed:
-            return self._container.mysql_router_service_enabled and not socket_file_exists
+            # After a snap refresh the service may be "enabled"
+            # without actually binding to the port
+            if not self._container.mysql_router_service_enabled or socket_file_exists:
+                return False
+            with socket.socket() as s:
+                s.settimeout(1)
+                return s.connect_ex(("localhost", self._charm._READ_WRITE_PORT)) == 0
         else:
             return socket_file_exists
 
--- a/common/common/workload.py
+++ b/common/common/workload.py
@@ -431,9 +431,11 @@ class RunningWorkload(Workload):
             # without actually binding to the port
             if not self._container.mysql_router_service_enabled or socket_file_exists:
                 return False
-            with socket.socket() as s:
-                s.settimeout(1)
-                return s.connect_ex(("localhost", self._charm._READ_WRITE_PORT)) == 0
+            try:
+                self._charm.wait_until_mysql_router_ready(event=event)
+            except AssertionError:
+                return False
+            return True
         else:
             return socket_file_exists
 
--- a/machines/src/charm.py
+++ b/machines/src/charm.py
@@ -24,6 +24,7 @@ import common.relations.cos
 import common.relations.database_provides
 import common.relations.database_requires
 import common.relations.tls
+import common.server_exceptions
 import common.workload
 import ops.log
 import tenacity
@@ -230,8 +231,10 @@ class MachineSubordinateRouterCharm(common.abstract_charm.MySQLRouterCharm):
                             with socket.socket(socket.AF_UNIX) as s:
                                 assert s.connect_ex(str(self._container.path(socket_file))) == 0
         except AssertionError:
-            logger.exception("Unable to connect to MySQL Router")
-            raise
+            logger.warning("Unable to connect to MySQL Router")
+            raise common.server_exceptions.Error(
+                ops.WaitingStatus("MySQL Router not ready")
+            ) from None
         else:
             logger.debug("MySQL Router is ready")
 
--- a/machines/src/snap.py
+++ b/machines/src/snap.py
@@ -179,7 +179,17 @@ class Snap(common.container.Container):
             if router_is_running:
                 self._snap.restart([self._SERVICE_NAME])
             else:
-                self._snap.start([self._SERVICE_NAME], enable=True)
+                # Retry snap start: after a snap refresh, the mysqlrouter service
+                # can crash repeatedly and hit systemd's StartLimitBurst limit.
+                # Waiting briefly and retrying lets the rate-limit window expire.
+                for attempt in tenacity.Retrying(
+                    stop=tenacity.stop_after_delay(60),
+                    wait=tenacity.wait_fixed(10),
+                    retry=tenacity.retry_if_exception_type(snap_lib.SnapError),
+                    reraise=True,
+                ):
+                    with attempt:
+                        self._snap.start([self._SERVICE_NAME], enable=True)
         else:
             self._snap.stop([self._SERVICE_NAME], disable=True)
 
--- a/machines/tests/integration/integration/test_upgrade.py
+++ b/machines/tests/integration/integration/test_upgrade.py
@@ -4,6 +4,7 @@
 import logging
 import shutil
 import zipfile
+from contextlib import suppress
 from pathlib import Path
 
 import jubilant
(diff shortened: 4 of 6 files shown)


astrojuanlu:
test_upgrade.py passing on machines on the first attempt:
- arm64 https://github.com/canonical/mysql-router-operators/actions/runs/32248231490/job/96081865551
- amd64 https://github.com/canonical/mysql-router-operators/actions/runs/32248231490/job/96081867487

astrojuanlu:
@sinclert-canonical @paulomach for your consideration. We don't have to merge this right now, but it's better than disabling the test, and looks like this fixes the issue.

astrojuanlu:
Too soon. Still flaky.

astrojuanlu:
Both machines `test_upgrade.py` passing on the 1st attempt.

common/common/workload.py
```diff
@@ -427,7 +427,13 @@ def _router_running_correctly(self, event) -> bool:
...
             return self._container.mysql_router_service_enabled
         elif charm_port_exposed:
-            return self._container.mysql_router_service_enabled and not socket_file_exists
+            # After a snap refresh the service may be "enabled"
+            # without actually binding to the port
+            if not self._container.mysql_router_service_enabled or socket_file_exists:
+                return False
+            with socket.socket() as s:
+                s.settimeout(1)
+                return s.connect_ex(("localhost", self._charm._READ_WRITE_PORT)) == 0
```
sinclert-canonical: This seems like an unrelated change to me, which I would rather leave outside of the PR.

---

However, if you want to preserve it, I would leverage the existing `_wait_until_router_ready` [method](https://github.com/canonical/mysql-router-operators/blob/efaf10416c225bfb1c76db202d99b80a22165940/machines/src/charm.py#L205-L236), which not only checks for the R/W port, but every port that is supposed to be open.

If you do this, you would need to wrap the whole check in a try-except block, and map the presence / absence of a raised exception into a boolean.

machines/src/snap.py
```diff
@@ -179,7 +179,17 @@ def update_mysql_router_service(self, *, enabled: bool, tls: bool = None) -> Non
...
+                # can crash repeatedly and hit systemd's StartLimitBurst limit.
+                # Waiting briefly and retrying lets the rate-limit window expire.
+                for attempt in tenacity.Retrying(
+                    stop=tenacity.stop_after_delay(60),
+                    wait=tenacity.wait_fixed(10),
+                    retry=tenacity.retry_if_exception_type(snap_lib.SnapError),
+                    reraise=True,
+                ):
+                    with attempt:
+                        self._snap.start([self._SERVICE_NAME], enable=True)
```
sinclert-canonical: As far as I understand this is the gist of it. Good finding!

machines/tests/integration/integration/test_upgrade.py
```diff
@@ -102,32 +104,49 @@ def test_upgrade_from_edge(juju: Juju, charm: str, continuous_writes) -> None:
     # Refresh will be incompatible on PR CI (not edge CI)
     # since unreleased charm versions are always marked as incompatible
     if router_status.current == "blocked" and "incompatible" in router_status.message:
-        logging.info("Application upgrade is blocked due to incompatibility")
+        with suppress(jubilant.TaskError):
```
sinclert-canonical: Do we need this suppression anymore?

machines/tests/integration/integration/test_upgrade.py
```diff
@@ -102,32 +104,49 @@ def test_upgrade_from_edge(juju: Juju, charm: str, continuous_writes) -> None:
...
-    )
+    # Wait for the upgrade to complete, resolving any units that error
+    # (e.g. the router was not ready within the 30-second timeout after a snap refresh)
+    # Resolving a unit makes Juju re-run the failed hook, which starts a fresh timeout window
+    def _resolve_errored_units(_) -> None:
+        for unit in get_app_units(juju, MYSQL_ROUTER_APP_NAME):
+            unit_status = juju.status().get_units(MYSQL_ROUTER_APP_NAME).get(unit)
+            if unit_status and unit_status.workload_status.current == "error":
+                logging.info(f"Unit {unit} is in error state, resolving and retrying")
+                juju.cli("resolved", unit)
```
sinclert-canonical: Is there any other wait to retry / resolve the enabling of the MySQL Router service? We never had to run the Juju resolve command to fix an application after a refresh, and I am not sure we should start recommending that to users.

The refresh functionality should work without it.

machines/src/snap.py
astrojuanlu: It's the first link of the chain, yes... 😅

common/common/workload.py
astrojuanlu: It was indeed needed to make the test pass. But I will look at this method, better to reuse it.

machines/tests/integration/integration/test_upgrade.py
astrojuanlu: > The refresh functionality should work without it.

That's fair enough. Extending the timeout is not bullet-proof though, so I'll see how can we make this robust at the charm level.

astrojuanlu: I think I fixed the reliability issues. Fingers crossed.

astrojuanlu:
This is ready for another round.

machines/src/charm.py
```diff
@@ -230,8 +231,10 @@ def wait_until_mysql_router_ready(self, *, event) -> None:
                             with socket.socket(socket.AF_UNIX) as s:
                                 assert s.connect_ex(str(self._container.path(socket_file))) == 0
         except AssertionError:
-            logger.exception("Unable to connect to MySQL Router")
-            raise
+            logger.warning("Unable to connect to MySQL Router")
+            raise common.server_exceptions.Error(
+                ops.WaitingStatus("MySQL Router not ready")
+            ) from None
```
sinclert-canonical: Any particular reason why to raise a custom error, instead of the assertion one?

sinclert-canonical (approved):
Thanks for looking into it!

astrojuanlu:
@paulomach called out that these tests are passing on the dpe branch 🤔 I'll have a closer look to see if we accidentally broke something on 8.4/edge

paulomach (approved):
Changes are adding resilience, so I am good with it. Just double check any diff with `dpe`

astrojuanlu:
I consider the investigation on the root cause mostly finished, see https://github.com/canonical/charmed-mysql-snap/issues/117. I started doing some work to properly demonstrate the RC fix, namely https://github.com/canonical/charmed-mysql-snap/pull/118, but that will be for another day. Even if we do decide to fix the root cause, these fixes improve resiliency.

machines/src/charm.py
astrojuanlu: This custom error has some special handling logic that takes care of setting the status, see

https://github.com/canonical/mysql-router-operators/blob/50d482e0b4446a1f613ff3ea9da294a55eb31aaa/common/common/abstract_charm.py#L383-L386

PR #212, opened 2026-08-27
[MISC] Migrate to ops-scenario 6.1 and ops 2.15

astrojuanlu:
So we can finally close https://github.com/canonical/mysql-router-operators/pull/143.

Marking it as bug fixed because of [CVE-2024-41129](https://nvd.nist.gov/vuln/detail/CVE-2024-41129).

### DIFF
--- a/common/common/relations/database_provides.py
+++ b/common/common/relations/database_provides.py
@@ -77,7 +77,7 @@ class _RelationThatRequestedUser(_Relation):
     ) -> None:
         super().__init__(relation=relation, interface=interface)
         self._interface = interface
-        if isinstance(event, ops.RelationBrokenEvent) and event.relation.id == self._id:
+        if not relation.active:
             raise _RelationBreaking
         self._database: str = self._databag["database"]
         if self._databag.get("extra-user-roles"):
@@ -146,7 +146,14 @@ class _RelationWithSharedUser(_Relation):
     ) -> None:
         super().__init__(relation=relation, interface=interface)
         self._interface = interface
-        self._local_databag = self._interface.fetch_my_relation_data([relation.id])[relation.id]
+        try:
+            self._local_databag = self._interface.fetch_my_relation_data([relation.id])[
+                relation.id
+            ]
+        except KeyError:
+            # Breaking relation excluded from data_interfaces lib (ops 2.10+ PR #1091).
+            # Access the local app databag directly from the relation object.
+            self._local_databag = relation.data[interface._model.app]
         for key in ("database", "username", "password", "endpoints", "read-only-endpoints"):
             if key not in self._local_databag:
                 raise _UserNotShared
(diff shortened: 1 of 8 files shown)


astrojuanlu:
Failure seems unrelated, I will look into it in another PR.

common/common/relations/database_provides.py
```diff
@@ -77,7 +77,7 @@ def __init__(
     ) -> None:
         super().__init__(relation=relation, interface=interface)
         self._interface = interface
-        if isinstance(event, ops.RelationBrokenEvent) and event.relation.id == self._id:
+        if not relation.active:
```
sinclert-canonical: Why not to preserve the `event.relation.id` check?

common/common/relations/database_provides.py
```diff
@@ -268,15 +287,16 @@ def reconcile_users(
                 _UnsupportedExtraUserRole,
             ):
                 pass
-        logger.debug(f"State of reconcile users {requested_users=}, {self._shared_users=}")
+        shared_users = self._shared_users(event)
```
sinclert-canonical: We are just caching the shared users here, correct?

sinclert-canonical (commented):
Further clarification about the steps taken to come up with this solution would be welcomed. Aside from the comments, I have additional questions:

- Why to include the _broken_ relationship (if any) every time we list _current_ relations? Is it to preserve backwards compatibility with the assumptions that the charm holds right now?

common/common/relations/database_provides.py
astrojuanlu: Indeed, reverted

common/common/relations/database_provides.py
astrojuanlu: Indeed, reverted

common/common/relations/database_provides.py
```diff
@@ -146,7 +146,9 @@ def __init__(
     ) -> None:
         super().__init__(relation=relation, interface=interface)
         self._interface = interface
-        self._local_databag = self._interface.fetch_my_relation_data([relation.id])[relation.id]
+        self._local_databag = self._interface.fetch_my_relation_data([relation.id]).get(
```
astrojuanlu: Breaking relation excluded from data_interfaces fetch methods (https://github.com/canonical/operator/pull/1091)

common/common/relations/database_requires.py
```diff
@@ -72,6 +72,8 @@ class CompleteConnectionInformation(ConnectionInformation):
     def __init__(self, *, interface: data_interfaces.DatabaseRequires, event) -> None:
         relations = interface.relations
         endpoint_name = interface.relation_name
+        if isinstance(event, ops.RelationBrokenEvent) and event.relation.name == endpoint_name:
```
astrojuanlu: Same as above

common/common/relations/remote_databag.py
```diff
@@ -30,7 +30,10 @@ def __init__(
         interface: data_interfaces.DatabaseRequires | data_interfaces.DatabaseProvides,
         relation: ops.Relation,
     ) -> None:
-        super().__init__(interface.fetch_relation_data()[relation.id])
+        super().__init__(
+            # fall back to the relation object for the remote app databag
+            interface.fetch_relation_data().get(relation.id, relation.data[relation.app])
```
astrojuanlu: Same as above

astrojuanlu:
> Further clarification about the steps taken to come up with this solution would be welcomed

First, I did some exploration to come up with the minimal version bump that would help us get rid of the flagged CVE. This turned out to be ops==2.15 and ops-scenario==6.1.

After that, I submitted the initial version you reviewed, which admittedly was a rough AI-generated patch.

I've now iterated a bit more to minimize the diff, including addressing the PR feedback.

sinclert-canonical (approved):
Thanks for the clarifications! Great patch indeed.

paulomach (approved):

